PR #12, opened 2025-01-23
Add java e2e tests

HarelM:
This should add e2e test that are based on Java to improve coverage.


HarelM:
There is room for improvement when it comes to waterway features that I need to understand, but otherwise, this looks good.

PR #22, opened 2025-02-04
Better support for highway types

### LINKED ISSUE #21: Support more highway types in search
An example of an interesting way only for search is the following:
https://www.openstreetmap.org/way/214778239
Other interesting ways might be:
`highway=cycleway|footway|path|track` that have a name.
Not sure how this will work with the mtb stuff, but let's hope that mtb only do not have a name.

HarelM:
- Resolves #21 

The following is the relevant search results after this change:
image


HarelM:
This looks very good in terms of coverage and adds the relevant support as far as I can tell.
More improvements can be made later on.

PR #35, opened 2025-02-20
Add more logic to finding the first point

### LINKED ISSUE #33: Wrong place for POI of Q relation
The following relation has a first correct way with a correct direction but the starting point of this relation is still in the wrong place:
https://israelhiking.osm.org.il/poi/OSM/relation_18299699?language=he

This is described here, but should be handled in this repo:
- https://github.com/IsraelHikingMap/Site/issues/2089

HarelM:
- Fixes #33


HarelM:
This is an overkill for this repo and is not needed at this point in time.
We can decide to use it in the future if need be.

PR #46, opened 2025-06-24
Add POIs for `tourism=alpine_hut`

zstadler:
Do we also want to add alpine huts to the POIs users can upload or update?

### DIFF
--- a/src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
+++ b/src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
@@ -383,7 +383,7 @@ public class PlanetSearchProfile implements Profile {
         return false;
     }
 
-    if (pointDocument.poiIcon == "icon-home" &&  pointDocument.poiCategory == "Wikipedia" && !feature.isPoint()) {
+    if (feature.getString("place") != null && pointDocument.poiCategory == "Wikipedia" && !feature.isPoint()) {
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -745,6 +745,11 @@ private void setIconColorCategory(PointDocument pointDocument, WithTags feature)
                 pointDocument.poiIcon = "icon-star";
                 pointDocument.poiCategory = "Other";
                 return;
+          case "alpine_hut":
+                pointDocument.poiIconColor = "#734a08";
+                pointDocument.poiIcon = "icon-home";
```
HarelM, line 750: Are we sure about this?
Not that I have a better idea which icon to use...

HarelM: Also, there's some special handling for icon-home here:
https://github.com/IsraelHikingMap/planet-search/blob/a767d9e8baf28d1a15abea21fef6939b5077bfe8/src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java#L386

HarelM (approved):

HarelM:
Let's see how this behaves. If there will be issues we can always fix them later on.

HarelM:
BTW, are there alpine_huts in Israel? I'm surprised by the fact the when using Israel this new code is passed...

PR #63, opened 2026-06-15
Index-time ranking signals (prominence, feature_class, area) + indexer resilience

sergeytkachenko:
## What & why

The companion index-side change to the Site relevance PR. It adds the build-time signals the new query-time ranking reads — **prominence**, **feature_class**, **area**, **population**, **alt_names**, an autocomplete **prefix** subfield, Hebrew normalization — and hardens the indexer against the silent document loss that made some results simply never appear.

> ### 🔗 Pairs with — and depends on — IsraelHikingMap/Site#2543
> This PR only **produces** the new index fields; the query-side PR
> **[IsraelHikingMap/Site#2543](https://github.com/IsraelHikingMap/Site/pull/2543)** is what actually
> *uses* them for ranking. **Merge Site#2543 first**, then this. Reasons:
> - On its own this PR ships index fields nothing reads yet — no user-visible effect until the Site PR is deployed.
> - Site#2543 is safe against today's (pre-reindex) index and is a reversible binary redeploy; this PR
>   only takes effect on the next whole-planet reindex (hours). Landing the cheap, reversible half first
>   lets you verify ranking on live traffic before committing to a reindex.
> - The Hebrew normalization here and Site#2543's query-side `query_norm` are two halves of one change —
>   they must both be deployed for Hebrew exact-name matching to stay correct.

**Takes effect on the next reindex.** The mapping changes are additive (new fields old queries ignore),
so the two PRs deploy independently and in either order without breakage — but the recommended order is
Site#2543 first (see box above).

## Scope

13 commits, 22 files (+3443/-102): 8 source files (`PlanetSearchProfile`, `ElasticsearchHelper`, `ProminenceCalculator`, `QRankIndex`, `OsmTagUtils`, `AccountingBulkListener`, `IndexingStats`, `PointDocument`, `MainClass`) + `pom.xml`, and ~12 test files. Highlights:

- **Prominence** — composite build-time importance from OSM tags + QRank (Wikimedia pageviews, optional `qrank.csv.gz`); clamped to `[0,1]`, floored so a query-time multiply never zeroes a hit. Pure/unit-tested.
- **Additive mapping** — `poiProminence`, `poiAreaNorm`, `intermittent`, `population`, multi-valued `alt_names`, `poiFeatureClass` (coarse OSM type incl. built/POI), and a `name.<lang>.prefix` edge-ngram subfield for as-you-type recall. Computed ranking fields carry a `poi*` prefix to mark them calculated, not raw OSM tags ([ADR-0027](https://github.com/IsraelHikingMap/planet-search/blob/mapeak-relevance-index/../../docs/adr/0027-remove-raw-prominence-component-fields.md) — raw prominence components dropped from the index).
- **Indexer resilience** — see below.
- **CLI** — whole-planet OSM source (`area=planet` → `aws:latest`); `--skip-tiles` collapses the tile pyramid to a z0 stub for a near-instant ES-only reindex (the search index is built identically); `--qrank-path` (empty ⇒ runs without QRank).

## Indexer resilience (the founding bug)

The original indexer swallowed bulk index errors silently, so a doc that failed to index just vanished from search. This PR makes indexing observable and lossless:

- **Accounting listener** (`AccountingBulkListener`) — every emitted op is counted exactly once as **indexed** or **failed**; the per-item walk handles short/partial bulk responses so no op can fall out of the `emitted == indexed + failed` invariant.
- **Bulk retry/backoff** — transient transport errors (timeouts, connection resets, 429/5xx) are retried with exponential backoff + full jitter on a **listener-owned executor**, so a backoff can't head-of-line-block ingest; the ingester is pinned to `noBackoff()` so it can never double-retry. Per-item 4xx (mapping/parse) are charged as genuine failures, never retried.
- **Ordered shutdown** — `finishIndexing()` flushes, closes the ingester, then drains offloaded retries (`awaitRetries()`) **before** the alias swap so no re-indexed doc races the swap.
- **Alias-swap accounting gate** — `MainClass` refuses to swap the alias unless `emitted == indexed + failed`; a mismatch (e.g. retries abandoned on a drain timeout) fails the build and leaves the freshly built index unpromoted for inspection rather than promoting a silently-short index. A partial swap (points promoted, bbox failed) is logged `SEVERE` naming the mismatched-alias state.

> Note: an earlier revision of this branch carried a two-bucket failure guard, a post-build reconcile gate, and a `--force-reindex`/`--min-protect-docs` reindex safety guard. Those were **removed** during review (commit `48f763b`) in favor of the simpler, exact `emitted == indexed + failed` accounting invariant above.

## ⚠️ One intended behavior change — Hebrew search

This mapping is additive for every language **except Hebrew**: `name.he` / `alt_names.he` move to a `hebrew_analyzer` with a doubled-matres fold (`וו→ו`, `יי→י`). After the reindex, Hebrew searches match spelling variants they previously missed. This is an intentional, requested recall improvement, **not a regression** — the Site PR's `query_norm` already mirrors this normalization so exact-name matching stays correct. Latin/other languages are byte-identical.

## Review fixes (two max-effort review passes)

Beyond the original scope, two rounds of max-effort review landed verified-defect fixes:

- **QRank multi-value `wikidata`** no longer drops the prominence signal (`Q42;Q64` splits, trims, takes max).
- **Retry no longer head-of-line-blocks ingest**; short-response leaks closed on both `afterBulk` paths; `emittedCount` counted only after a successful `add()`.
- **`ProminenceCalculator.clamp01` sanitizes `NaN`**; negative elevations honored; non-icon emit path now runs the ranking signals.
- **Alias-swap accounting gate** + partial-swap `SEVERE` logging (above).
- **Population hardening** — a garbage tag that overflows to `+Infinity` no longer clamps a settlement to `Integer.MAX_VALUE` (`Double.isFinite` guard); non-settlement `place=*` values stay `null` (ES `missing:1.0`) instead of a fabricated `20`. Logic extracted to the pure `OsmTagUtils.computePopulation`.

Full per-fix notes (commit SHAs + ADR refs) are in the repo `CHANGELOG.md`.

## Tests & build

**146 unit tests pass** on JDK 21 (run in `maven:3.9-eclipse-temurin-21`; `mvn test` runs the `@Tag("unit")` set). ES-free unit coverage: prominence (`ProminenceCalculatorTest`, `InsertProminenceFloorTest`, `ParseFirstNumberTest`), QRank (`QRankIndexTest`), feature_class (`SetFeatureClassTest`), population (`ComputePopulationTest`), alt_names + Hebrew matres (`AddAltNamesTest`, `HebrewMatresRuleTest`), index mapping (`PointsIndexMappingTest`), alias swap (`SwitchAliasTest`, `ElasticsearchHelperTest`), and indexer resilience (`IndexerResilienceTest`, `BulkIngesterAccountingTest`). End-to-end (`E2ETest`) runs under the `run-all-tests` profile against a Testcontainers ES.

## Deploy

Merge after Site#2543 is live. On the next whole-planet rebuild the indexer writes to the inactive index, drains retries, restores search settings, then **atomically swaps the alias only if the accounting invariant holds** (otherwise it fails the build and leaves the new index unpromoted). Build time is ~unchanged; `--skip-tiles` is available for faster ES-only reindex turnarounds.


src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -20,13 +20,154 @@ public class ElasticsearchHelper {
...
+    return input
+        .replaceAll("וו", HEBREW_VAV)
+        .replaceAll("יי", HEBREW_YOD);
+  }
+
   public static ElasticsearchClient createElasticsearchClient(String esAddress) {
     Logger.getLogger("org.elasticsearch.client.RestClient").setLevel(Level.OFF);
-    RestClient restClient = RestClient.builder(HttpHost.create(esAddress)).build();
+    RestClient restClient = RestClient.builder(HttpHost.create(esAddress))
+        // Step D — indexer resilience: the default per-request socket timeout is 30s,
```
HarelM: Move in method comments to be method declaration comments instead (i.e. move them above the method name), and also reduce verbosity, make the comments shorter.

HarelM:
The comments are too verbose to read through this. I'm having a hard time understanding what's important and what's no from the comments. Please tone it down a lot.

HarelM:
Also in general, I would prefer, if possible to do it the other way around.
Merge this to update planet-search, allow it to run on production and generate the new index, see that it's not taking a lot longer than before, and then update site to use the new index.
If this is not possible let me know.
Can you clarify the `--force-reindex` flag? I'm not sure I understand it.
In general, the default should be to create a new index and swap, if there's a need to do something different for debug, then it shouldn't be the default.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -86,6 +432,95 @@ static private final void CoalesceIntoMap(Map<String, String> map, String langua
...
+  /**
+   * PURE (no I/O, no {@code WithTags}) builder for the {@code alt_names} map, so the {@code ;}-split,
+   * trim, de-dup and old_name-exclusion logic is directly unit-testable (mirrors the
+   * {@code flooredProminence} extraction pattern). Returns {@code null} when no variant tags are
+   * present (so {@code addAltNames} leaves {@code alt_names} null and it is omitted from JSON).
+   *
+   * @param supportedLanguages the per-language keys to look up suffixed tags for
+   * @param tagLookup          maps an OSM tag key to its raw value (null if absent)
+   */
+  static Map<String, List<String>> buildAltNames(String[] supportedLanguages,
```
HarelM: I don't think this is used outside this class.
If you are having a hard time testing it, I suggest move it to a different class and have that tested there. Maybe a utilities class.

pom.xml
```diff
@@ -37,6 +37,12 @@
       <scope>test</scope>
     </dependency>
 
+    <!--
```
HarelM: This comment is weird... I suggest removing it.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -27,14 +31,66 @@
...
+  // Empty index when no QRank file is provided (local builds) — lookups return 0.
+  private final QRankIndex qrankIndex;
+
+  // Bulk indexer: thread-safe, buffers operations and flushes them in batches.
+  // Planetiler emits features from multiple worker threads, so BulkIngester
+  // (which handles concurrent add() internally) is the right tool rather than
+  // manually batching BulkRequests with our own locking.
+  private final BulkIngester<Void> bulkIngester;
+  // Counters so indexing is observable instead of failing silently.
+  private final LongAdder indexedCount = new LongAdder();
```
HarelM: Move all index count stuff to either the elastic helper or a different class, this in not the focus of this class and should be moved from here.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -44,11 +100,301 @@ public class PlanetSearchProfile implements Profile {
 
   public PlanetSearchProfile(PlanetilerConfig config, ElasticsearchClient esClient, String pointsIndexName,
       String bboxIndexName, String[] supportedLnaguages) {
+    this(config, esClient, pointsIndexName, bboxIndexName, supportedLnaguages, QRankIndex.empty());
+  }
+
+  public PlanetSearchProfile(PlanetilerConfig config, ElasticsearchClient esClient, String pointsIndexName,
```
HarelM, line 55: Why two constructors are needed?

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -44,11 +100,301 @@ public class PlanetSearchProfile implements Profile {
...
+  }
+
+  /**
+   * BulkListener that surfaces and counts every bulk outcome instead of
+   * swallowing errors. Extracted from an anonymous inner class so the per-item
+   * classification (indexed vs failed) and the whole-batch failure path are
+   * unit-testable by driving {@code afterBulk} directly. The classification
+   * logic is unchanged from the original anonymous listener.
+   */
+  static final class AccountingBulkListener implements BulkListener<Void> {
```
HarelM: Move this to a different file.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -86,6 +432,95 @@ static private final void CoalesceIntoMap(Map<String, String> map, String langua
     }
   }
 
+  /**
+   * Language-neutral OSM variant-name tag bases (no {@code :<lang>} suffix). These also have
+   * per-language forms ({@code alt_name:he}, {@code official_name:en}, ...). {@code old_name} is
+   * DELIBERATELY EXCLUDED — it is a documented Nominatim noise source (a stale former name can
+   * overtake the real one). See ADR-0011.
+   */
+  static private final String[] ALT_NAME_TAG_BASES = {
```
HarelM: Move this to the begging of the class.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -97,11 +532,334 @@ private void convertTagsToDocument(PointDocument pointDocument, WithTags feature
...
+   * primary OSM type tags and store it on {@code pointDocument.feature_class}. This is the kind-of-
+   * thing signal the engine needs to rank "Mount Shasta" as a SUMMIT, not a homonym hotel — distinct
+   * from poiCategory ("Wikipedia"/"Other"), which never identifies the feature type.
+   *
+   * <p>Reads the same tags {@link #setProminence} already consults (no new tag access). Most-specific
+   * tag wins; leaves the field null when nothing recognised is present (omitted from JSON, so the
+   * ~22.8M docs without a known type carry no extra bytes). Normalises OSM values into a small,
+   * stable vocabulary so the query side can match on a known set rather than raw OSM strings.
+   */
+  void setFeatureClass(PointDocument pointDocument, WithTags feature) {
```
HarelM: This is a one line method, simply inline it when needed.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -97,11 +532,334 @@ private void convertTagsToDocument(PointDocument pointDocument, WithTags feature
...
+    pointDocument.feature_class = classifyFeatureClass(feature::getString);
+  }
+
+  /**
+   * PURE (no {@code WithTags}) core of {@link #setFeatureClass}, so the OSM-tag -> feature_class
+   * switch is directly unit-testable from a literal tag map (mirrors {@code buildAltNames} /
+   * {@code flooredProminence}). Returns null when nothing recognised is present.
+   */
+  static String classifyFeatureClass(java.util.function.Function<String, String> tagLookup) {
```
HarelM: This looks very similar to the icon code, I don't want this duplicated, so if this can leverage the tags to icon map it would be a lot better.

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -1,10 +1,24 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
 class PointDocument {
   public Map<String, String> name = new HashMap<String, String>();
+  /**
+   * ADR-0011 item 1 — variant/alternative names (alt_name, alt_name:&lt;lang&gt;, official_name,
+   * short_name, loc_name, int_name, ...) keyed by language. SEPARATE from {@link #name} on purpose
+   * (folding variants into name would break ranking/display/ADR-0009). Searched at strictly lower
+   * boosts than name and never read by display (HitToFeature keeps the canonical name).
+   *
+   * <p>Left null until at least one variant tag is present (NON_NULL on the class then omits it from
```
HarelM: See if you can avoid html tags in the comments. I don't see this help reading it.
Also remove all the ADR-00XX comments. I have no need to track those.

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -18,4 +32,37 @@ class PointDocument {
   public double poiLength = 0;
   public String website;
   public double[] location;
+
+  // --- Ranking signals (added for search relevance; null => omitted from JSON, ES uses missing:1.0) ---
+  /** Composite prominence score in [0,1], floored >0 so a query-time multiply never zeroes. Hot path. */
+  public Float prominence;
+  /** Raw components, stored index:false for re-tuning weights without a reindex. */
```
HarelM: If this is used only for debug, please remove them, I prefer the document to be as simple as possible.

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -18,4 +32,37 @@ class PointDocument {
...
+  /** Population — place/admin layer only (city/town/village...); null for POIs. */
+  public Integer population;
+  /**
+   * Coarse FEATURE TYPE derived from the primary OSM type tags (natural/place/waterway/...), e.g.
+   * "peak", "lake", "waterfall", "city", "spring". Distinct from {@link #poiCategory}, which is an
+   * app/source bucket ("Wikipedia"/"Hiking"/"Other") and does NOT identify the kind of feature.
+   * Used as a query-time ranking signal (boost docs whose class matches the query's intent) — see
+   * the rescore scoring model. Null when no recognised type tag is present (omitted from JSON).
+   */
+  public String feature_class;
```
HarelM: use `poi` prefix for things that are not osm tags, see above with length, category etc.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -129,4 +373,115 @@ public static void switchAlias(ElasticsearchClient esClient, String indexAlias,
...
+   *
+   * <p>Because re-emitting the same id overwrites in place, the comparison is made
+   * against {@code count + deleted} (the number of distinct index ops that landed),
+   * not the bare live {@code count} — otherwise legitimate dedup overwrites would
+   * look like loss. The build fails when the landed count is short of expected by
+   * more than {@code shortfallTolerance} (a fraction, e.g. 0.001 for 0.1%).
+   *
+   * @return null when within tolerance; otherwise a human-readable failure message.
+   */
+  public static String reconcileLivePoints(long emittedPoints, long genuineDataFailures,
```
HarelM: Some of the code here looks like a debug tool to understand if an index is OK or not. I'm guessing you found the root cause (missing bulk index and proper handling of failures) and fixed it, so I'm not sure it's worth keeping all this code. I think it complicates things without good value.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -37,31 +178,130 @@ public static String createPointsIndex(ElasticsearchClient esClient, String inde
...
+                            .normalizer(isHebrew ? "hebrew_normalizer" : "universal_normalizer")))
+                    .fields("prefix", f -> f
+                        .text(pt -> pt
+                            .analyzer("prefix_index_analyzer")
+                            .searchAnalyzer("prefix_search_analyzer")))));
+            // alt_names.<lang> — SEPARATE demoted variant-name field (ADR-0011 item 1). NOT folded
+            // into name (folding breaks ranking/display/ADR-0009). Same analyzer choice as name so
+            // he variants get the matres collapse too. No .prefix subfield (alt prefix is out of
+            // this bundle's query design).
+            m.properties("alt_names." + lang, k -> k
```
HarelM, line 137: This is a list of names, doesn't elastic treat it as a "sentence" instead of different words or phrases?

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -37,31 +178,130 @@ public static String createPointsIndex(ElasticsearchClient esClient, String inde
...
+          // Raw components kept for re-tuning weights without a reindex: not searchable
+          // (index:false) but doc_values stay on so they remain readable.
+          m.properties("prom_base", n -> n.float_(f -> f.index(false)));
+          m.properties("prom_qrank_norm", n -> n.float_(f -> f.index(false)));
+          m.properties("prom_meta", n -> n.float_(f -> f.index(false)));
+          m.properties("ele_norm", n -> n.float_(f -> f.index(false)));
+          m.properties("qrank_raw", n -> n.long_(f -> f.index(false)));
+          // ADR-0014 enrichment signals (all index:false — query-time scoring inputs, re-weighted
+          // in the script without a reindex; doc_values stay on so the script can read them).
+          m.properties("area_norm", n -> n.float_(f -> f.index(false)));
```
HarelM: Is there any change related to incorporating bbox index into this index to improve search performance and reduce the extra call for each search restuls to bbox index to get the "container" name?

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Done in a38eb2a — `AccountingBulkListener` is now its own file.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Done in a38eb2a — moved to a new `IndexingStats` class.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Moved to a new `OsmTagUtils` utility class and tested there (a38eb2a).

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Rather than reorder it, I pulled it out of the class entirely into `OsmTagUtils` (a38eb2a) — so the "top of class" move is moot, there's nothing left here to reposition. Happy to adjust its placement within the new file if you like.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Removed — only one constructor now (a38eb2a).

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Inlined; `setFeatureClass` is gone (a38eb2a).

src/main/java/il/org/osm/israelhiking/PointDocument.java
sergeytkachenko: Removed — all raw component fields are gone now (a38eb2a).

pom.xml
sergeytkachenko: Trimmed to a single line in a38eb2a — kept one line because the 8.x-client-against-7.17-server skew is intentional and non-obvious (it's what enables the BulkIngester path), so a future reader shouldn't "fix" it. Can drop it entirely if you'd still rather have nothing.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
sergeytkachenko: Trimmed and moved to the method declaration in a38eb2a. I'll do another pass to shorten the rest and drop the ADR-00xx references across the diff.

src/main/java/il/org/osm/israelhiking/PointDocument.java
sergeytkachenko: Removed the HTML tags and the ADR-00xx references on this file (a38eb2a). Will sweep the remaining files too.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Good catch — you're right these are two tag→string mappings over the same keys. Worth noting the expensive half is already shared: both paths take the same `OsmTagCache`, so the ~21 tag reads happen once per feature and feed both `setIconColorCategory` and `classifyFeatureClass`. What's still duplicated is the mapping table itself.

I didn't fold them into one table here because the two consumers have different return shapes: `setIconColorCategory` is void and writes three fields (icon + color + category), while `classifyFeatureClass` returns a single coarse class String. Unifying means a shared row type (`class → {icon, color, category}`) and reworking both switches to read from it — a real refactor, not a mechanical merge.

My recommendation: keep this PR scoped to the ranking signals and do the table unification as a follow-up — I'll open an issue and link it. Happy to fold it in here instead if you'd rather not split it.


Tracking: #64.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
sergeytkachenko: Good eye — part of this is diagnostics, but the gate itself earns its keep, so let me separate the two.

The gate is an end-to-end correctness backstop: after the build it reads the live points alias doc-stats and compares `emitted − genuineDataFailures` against `count + deleted` (distinct ops landed, so legitimate dedup overwrites aren't counted as loss), failing the build if the shortfall exceeds 0.1%. That's deliberately independent of the per-item failure counter (`hasIndexingFailures`), which only reflects what the client *observed* — the gate cross-checks the actual live index, so it catches silent loss the per-item path tolerates (e.g. transient whole-batch charges treated as phantom timeouts). It also fails *open* on a stats-read hiccup, so a flaky probe can't break an otherwise-good build.

One correction to be precise: the alias swap happens *before* the gate runs, so it doesn't stop a partial index from going live — it fails the build loudly so an operator rolls back. (The retry/backoff you may be thinking of lives in `AccountingBulkListener`, not here.)

The verbose failure banner is the part that's really just diagnostics — happy to slim that down while keeping the gate logic. Want me to?

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
sergeytkachenko: Good question — array handling is exactly what prevents the "sentence" behavior. `alt_names.<lang>` is a `text` field populated from a `List<String>` (each variant split on `;`, trimmed, de-duped). Elastic analyzes each array element independently and inserts a `position_increment_gap` between elements — we don't set it, so the ES default of 100 applies. That gap is the safeguard: it makes phrase/proximity queries unable to match across element boundaries, so two adjacent variants can't be matched as one phrase. Each variant stays its own searchable phrase rather than being folded into a run-on string.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
sergeytkachenko: Out of scope here — this PR only adds index-time ranking-signal mappings to the points index (`area_norm`, `intermittent`, etc.), not the points/bbox topology. Folding the container in would remove the per-result bbox lookup you're describing, but it's a separate change that needs its own design and a before/after benchmark on that hot path. I'll open a follow-up issue to track it and link it here.


Tracking: #65.

sergeytkachenko:
Thanks for the review. Summary of what changed since you looked (head is now \`6ab67ba\`):

- Extracted \`AccountingBulkListener\`, \`IndexingStats\` and \`OsmTagUtils\` out of \`PlanetSearchProfile\`; removed the second constructor; inlined \`setFeatureClass\`; removed all the raw prominence-component fields from the document (it's simpler now). Trimmed comments and started dropping the ADR-00xx references — I'll finish that sweep.
- Still open per the inline threads: unifying classify with the icon map, and whether to rename \`feature_class\` -> \`poiFeatureClass\` (that one touches the Site repo).

On `--force-reindex` and the order: the default already **is** create-a-new-index-and-swap the alias — every run builds a fresh index and only swaps it in at the end. `--force-reindex` is purely a safety override: without it, the indexer refuses to overwrite an existing **populated, live** index (this guard exists because a stray run once wiped the live planet index down to a regional extract). Normal/CI runs need nothing; you only pass `--force-reindex` for an intentional full rebuild over live data. No debug-only default behavior.

On sequencing — fully agree, that's the safer order: merge + deploy planet-search, let it build the new index on prod, confirm the build time is acceptable, **then** update Site to read the new fields. The Site PR is additive (missing fields just fall back to neutral), so it can wait until the index is live. I'll hold the Site side until you've confirmed the prod reindex.

HarelM:
Ok, good to hear we can start with this PR and only after the index is live move to updating the site.
Yes, please change poiFeatureClass instead of feature class, and prefix all non OSM tags with `poiXXX`. This way it's easy to understand which part is calculated and which part came straight from OSM.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
HarelM: This is because it is mapped as a `keyword` and not the default `text`, right?

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
HarelM: I don't think a gate is needed.
Also there shouldn't be delete count or updates as all documents should be unique.
Instead of all this, just log at the end of the run a single line with statistics - how many documents were indexed vs how many documents were supposed to be indexed. I'll consider adding an alarm to cloudwatch if the diff was too big, but that's outside the scope of this repo. It should only report to log the stats.

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+      });
+    } else {
+      indexedCount.add(response.items().size());
+    }
+  }
+
+  @Override
+  public void afterBulk(long executionId, BulkRequest request, List<Void> contexts, Throwable failure) {
+    // A whole-batch failure has no per-item breakdown, and the 8.x BulkIngester does not
```
HarelM: Still too many in method comment... move them to the method dicumentation.
There are more comments code than actual code...

HarelM: I'll let you finish the sweep to remove comments.
ping me when this is ready for another round of review.

sergeytkachenko:
Ready for another round — comment sweep is done.

**Since the last review (head `8b82a79`):**
- Swept all in-method comments to method-level docs; removed every HTML tag and ADR-00xx reference across `src/main`.
- `AccountingBulkListener`: retries no longer run inline on the bulk scheduler thread (they were head-of-line-blocking ingest under transient ES errors) — they're offloaded to a small pool and drained before the alias swap. Folded the two duplicate per-op classification loops into one `classifyAndCount`.
- `QRankIndex`: multi-value `wikidata` (`Q42;Q64`) now contributes its QRank instead of being dropped.
- Minor cleanups: removed two unused constructors, moved the test-only executor out of prod, tightened a few stale comments.

**On `alt_names` (your question):** it's mapped as `text` (analyzed), with a `.keyword` sub-field for exact matching — not `keyword`. Each array element is analyzed independently, so adjacent variants aren't matched as one phrase.

**One open item:** `classifyFeatureClass` and `setIconColorCategory` still hold two switch tables over the same OSM keys. I'd prefer to unify them onto a shared `class → {icon, color, category}` table as a separate follow-up (left a `TODO(#64)` anchor on both) to keep this PR scoped to the ranking signals — ok with you, or fold it in here?

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
sergeytkachenko: Not quite — the safeguard comes from the base `text` mapping, not `keyword`. `alt_names.<lang>` is `text` (analyzed); ES analyzes each array element independently with a `position_increment_gap` between them (default 100), and that gap is what stops a phrase query matching across two adjacent variants. A `keyword` field wouldn't analyze at all, so it'd be exact-whole-string only — no per-word recall.

The `.keyword` here is a sub-field on top of the `text` field, and it's used: the Site exact-alt tier does a phrase match on `alt_names.<l>.keyword` (ConstantScore boost 6) so a query that exactly equals a variant (e.g. "IBT") surfaces the feature but ranks below a primary-name match (boost 12). So the field carries both: `text` for analyzed recall + the gap safeguard, `.keyword` for the exact-variant tier.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
sergeytkachenko: Done — dropped the gate entirely, along with the delete/update counts and the two-bucket accounting, in 48f763b (`reconcileLivePoints`, `getLiveAliasDocsStats`, `assertSafeToReindex` are all gone; -174 lines in this file). The indexer now just logs a single line at the end: `Indexing finished: indexed N of M emitted document(s) (F failed).` CloudWatch alarming on the diff left out of this repo, as you said.

src/main/java/il/org/osm/israelhiking/PointDocument.java
sergeytkachenko: Done — `feature_class` → `poiFeatureClass`, and every computed field now carries the `poi` prefix: `poiProminence`, `poiAreaNorm`, `poiFeatureClass`, plus the existing `poiCategory`/`poiIcon`/`poiIconColor`/`poiSource`/`poiDifficulty`/`poiLength`. Raw OSM-tag / intrinsic fields keep their names (`name`, `alt_names`, `description`, `wikidata`, `image`, `wikimedia_commons`, `website`, `population`, `intermittent`, `location`).

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
sergeytkachenko: Comment sweep done — the explanations are now in method-level javadoc, and the in-method `//` notes left are short, line-specific ones. `AccountingBulkListener` is down to ~0.24 comments-to-code; no more "more comments than code". Ready for another round whenever you have a moment.

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,268 @@
...
+
+  private static ExecutorService defaultRetryExecutor() {
+    return Executors.newFixedThreadPool(RETRY_POOL_SIZE, r -> {
+      Thread t = new Thread(r, "es-bulk-retry");
+      t.setDaemon(true);
+      return t;
+    });
+  }
+
+  private void recordFailure() {
```
HarelM: Remove one line method.

HarelM:
I went over the first two files. 
Comments are still too verbose.
Please only leave comments when things are not self explained by the code. 
For a well written code, a good name of a method or a variable should be enough to completely avoid the need for a comment.

HarelM:
I've also saw a fair share of code aimed for unit tests, which is great, but it shouldn't be xjust for unit tests" it should be part of the production code that happens to be used by unit tests as well. 
So if there's two constructors for a class, one used by production and the other by unit tests, that not a good practice. 
If a class has a method that is only being using in tests, that's not good either. 
Ping me when this is ready for another review.

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
sergeytkachenko: Done — removed in `eb3078f`. `recordFailure()` is gone; its body (`failedCount.increment()`) is inlined at the call sites (the whole-batch non-retryable path, the per-op classifier, and the retry-exhausted/non-retryable-resubmit paths).

sergeytkachenko:
@HarelM — both of your comments are addressed in `eb3078f`. Answering them directly:

**1. Comments too verbose (a good name should remove the need for a comment).** Done across *all* the changed files, not just the first two. I removed every comment that only restates what a well-named field/method/constant already says (the `// header: Entity,QRank`-style WHAT notes, the field-by-field restatements in `PointDocument`, the analyzer-name restatements, etc.) and kept only the rationale a name can't carry — e.g. why the Hebrew matres fold is doubled-only (avoids merging ~7 real homographs), why `noBackoff()` is pinned, why the prominence floor / NaN-guard exist. `AccountingBulkListener` in particular went from "more comments than code" to method-level javadoc + a few short line-specific notes.

**2. Code aimed only at unit tests → production code the tests happen to drive.**
- `AccountingBulkListener` is back to a **single** constructor `(ElasticsearchClient, IndexingStats)`. The test-only constructor, the `Sleeper` interface, and the `TestExecutors` helper are gone. The listener owns its real retry pool and sleeps via `Thread.sleep`.
- The retry loop is now `resubmitWithBackoff(executionId, ops, BulkAttempt, BackoffSleep)` — its two effects (submit-a-bulk, sleep-between-attempts) are parameters. Production wires them to `esClient.bulk(...)` / `Thread::sleep`; the unit tests call the **same** method with a stubbed attempt-sequence and a no-op sleep. Nothing exists solely for tests — they exercise the real loop.
- `ProminenceCalculator.compute(...)` now returns a bare `float`. I dropped the `Result` holder whose `base/qrankNorm/meta/eleNorm/qrankRaw` fields were there only so the test could assert on intermediate math. The test now asserts on the observable score (e.g. a below-sea-level peak scoring equal to a sea-level peak proves the elevation clamp without reading an internal field).
- The one-line `recordFailure()` you flagged is inlined to `failedCount.increment()` at all call sites; the method is removed.

Unit suite is green on JDK 21 (142 tests).

One trade-off to flag: dropping the injected sleep/executor means the `awaitRetries()` **drain-ordering** check (which needed that seam to park a retry mid-flight) is no longer a unit test — that ordering is now covered by the live-ES `E2ETest` only. The retry *state machine* itself (multi-attempt recovery, the 5-attempt cap, per-item-503 handling, surplus-op-not-dropped) stays fully unit-tested via the parameterized loop. Happy to add it back in another shape if you'd prefer it unit-covered.

Ready for another round whenever you have a moment. 🙏

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,261 @@
...
+import co.elastic.clients.elasticsearch.core.BulkRequest;
+import co.elastic.clients.elasticsearch.core.BulkResponse;
+import co.elastic.clients.elasticsearch.core.bulk.BulkOperation;
+import co.elastic.clients.elasticsearch.core.bulk.BulkResponseItem;
+
+/**
+ * BulkListener that surfaces and counts every bulk outcome instead of swallowing errors.
+ */
+final class AccountingBulkListener implements BulkListener<Void> {
+  // Log under the profile's logger so the indexing diagnostics stay in one place (and tests that
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,261 @@
...
+  // assert on the PlanetSearchProfile logger still see them).
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,261 @@
...
+  private static final Logger LOGGER = Logger.getLogger(PlanetSearchProfile.class.getName());
+
+  // Retry/backoff tuning for the whole-batch transient path.
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,261 @@
...
+  static final int MAX_RETRY_ATTEMPTS = 5;          // total resubmits (1s,2s,4s,8s,16s)
+  static final long BASE_BACKOFF_MILLIS = 1_000L;   // first backoff
+  static final long MAX_BACKOFF_MILLIS = 16_000L;   // cap per attempt
+
+  static final int RETRY_POOL_SIZE = 4;
+  // One batch's retries are bounded (~5 attempts x 16s backoff + ES socket timeouts ~= a few minutes);
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/AccountingBulkListener.java
```diff
@@ -0,0 +1,261 @@
...
+  // 30 min leaves ample room for a queue of batches draining on RETRY_POOL_SIZE threads at shutdown.
```
HarelM:
```suggestion
```

HarelM:
@sergeytkachenko it feels like you are handing off my comments to an AI without checking what it did first as I keep repeating them.
This is consuming my time.
Please look at the review notes, do the changes however you want, but please review them before pinging me.

src/main/java/il/org/osm/israelhiking/QRankIndex.java
```diff
@@ -0,0 +1,123 @@
...
+   * Load a gzipped qrank.csv.gz. If csvGz is null, missing, or empty, returns an empty index (no
+   * exception) so local builds can run without the file.
+   */
+  static QRankIndex load(Path csvGz) {
+    if (csvGz == null || !Files.isRegularFile(csvGz)) {
+      LOGGER.info("QRankIndex: no QRank file provided — running with an empty index (lookups return 0)");
+      return empty();
+    }
+    long startMs = System.currentTimeMillis();
+    LongIntHashMap map = new LongIntHashMap(32_000_000);
```
HarelM, line 48: Why 32M? What will happen if the file gets larger?

src/main/java/il/org/osm/israelhiking/QRankIndex.java
```diff
@@ -0,0 +1,123 @@
...
+        BufferedReader br = new BufferedReader(new InputStreamReader(gz, StandardCharsets.UTF_8), 1 << 20)) {
+      String line = br.readLine();
+      while ((line = br.readLine()) != null) {
+        int comma = line.indexOf(',');
+        if (comma <= 1 || line.charAt(0) != 'Q') {
+          continue;
+        }
+        try {
+          long qid = Long.parseLong(line, 1, comma, 10); // skip the leading 'Q', no substring alloc
+          int qrank = Integer.parseInt(line, comma + 1, line.length(), 10);
```
HarelM, line 61: According to the following, there should be a label parsing, shouldn't there?
https://github.com/brawer/wikidata-qrank/blob/main/README.md

If the above is true, I would recommend maybe using a CSV parsing library instead of writing the CSV parsing code here, tempting as it may be to write CSV parsing code...

src/main/java/il/org/osm/israelhiking/QRankIndex.java
```diff
@@ -0,0 +1,123 @@
...
+    if (wikidata == null) {
+      return 0;
+    }
+    if (wikidata.indexOf(';') < 0) {
+      return qrankFor(wikidata.trim()); // single value — no split allocation
+    }
+    // Manual indexOf walk over the ';'-separated parts (no regex/array alloc); empty parts give 0.
+    long best = 0;
+    int start = 0;
+    while (start <= wikidata.length()) {
```
HarelM, line 93: Why using indexing for string? Why not use string split? and make this code a lot simpler.
You might also use stream to find the minimum value.
This is very verbose code...

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Computes a composite prominence score in [0,1], used at query time as a field_value_factor
+ * multiplier so a major peak / well-known place outranks an obscure node with the same name.
+ *
+ * Formula: prominence = clamp01( 0.05 + 0.45*base + 0.40*qnorm + 0.10*meta )
+ *   - floor 0.05: never zero, so a query-time multiply never annihilates a hit.
+ *   - base: feature-class prior (elevation-scaled peaks, place hierarchy, parks, water, baseline).
+ *   - qnorm: log-normalized QRank (Wikimedia pageviews), 0 when no wikidata/QRank hit.
+ *   - meta: richness from image / website / wikidata tags.
```
HarelM, line 11: Is wikidata calculated "twice"? According to what is explained here, it is also taken into account in qnorm, isn't it?

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Pure and side-effect-free so it is unit-testable without planetiler/ES.
+ */
+final class ProminenceCalculator {
+
+  static final double QRANK_REF = 3_000_000.0;
+  static final double MAX_ELE_M = 8849.0;
+  /** A 0 would zero out the query-time multiply, so prominence is floored, never zero. */
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,88 @@
...
+    double base = baseScore(naturalTag, placeTag, boundaryTag, tourismTag, historicTag, waterwayTag, eleNorm);
+
+    double qnorm = (qrankRaw <= 1) ? 0.0 : clamp01(Math.log(qrankRaw) / Math.log(QRANK_REF));
+
+    double meta = clamp01(0.40 * (hasImage ? 1 : 0) + 0.35 * (hasWebsite ? 1 : 0) + 0.25 * (hasWikidata ? 1 : 0));
+
+    return (float) clamp01(FLOOR + 0.45 * base + 0.40 * qnorm + 0.10 * meta);
+  }
+
+  /** Feature-class prior in [0,1]. Highest signal wins (a place that is also a peak gets the peak rule). */
```
HarelM, line 51: Does prior means priority?
If so, let's not abbreviate words...

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,88 @@
...
+  static float compute(String naturalTag, String placeTag, String boundaryTag, String tourismTag,
+      String historicTag, String waterwayTag, double ele, boolean hasImage, boolean hasWebsite,
+      boolean hasWikidata, long qrankRaw) {
+
+    double eleNorm = Double.isNaN(ele) ? 0.0 : clamp01(Math.log1p(Math.max(0, ele)) / Math.log1p(MAX_ELE_M));
+    double base = baseScore(naturalTag, placeTag, boundaryTag, tourismTag, historicTag, waterwayTag, eleNorm);
+
+    double qnorm = (qrankRaw <= 1) ? 0.0 : clamp01(Math.log(qrankRaw) / Math.log(QRANK_REF));
+
+    double meta = clamp01(0.40 * (hasImage ? 1 : 0) + 0.35 * (hasWebsite ? 1 : 0) + 0.25 * (hasWikidata ? 1 : 0));
```
HarelM, line 46: Why is clamp01 needed here?

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,88 @@
...
+      return 0.55;
+    }
+    if (historic != null) {
+      return 0.55; // ruins, archaeological_site, monument, memorial, castle...
+    }
+    if ("spring".equals(natural) || "hot_spring".equals(natural) || "cave_entrance".equals(natural)
+        || waterway != null) {
+      return 0.30;
+    }
+    return 0.25; // generic named node baseline
```
HarelM, line 79: There are other "interesting" points we are currently "highlighting" in our app, such as monuments, waterfalls, hiking/biking routes etc.
Should this code take that into consideration?
I think a better approach would be to remove this method completely and use iconcolorcategory and add the factor calculation there, otherwise, it won't be possible to maintain these together...

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -14,8 +20,17 @@ class PointDocument {
...
+
+  // Computed ranking signals (null => omitted from JSON, ES uses missing:1.0). The poi* prefix marks
+  // them as calculated, not raw OSM tags.
+  public Float poiProminence;
+  // A size proxy (a large lake outranks a same-named pond); set only for polygons.
+  public Float poiAreaNorm;
+  public Boolean intermittent;
+  public Integer population;
+  // Distinct from poiCategory (an app/source bucket); this is the OSM feature type ("peak"/"lake"/...).
+  public String poiFeatureClass;
```
HarelM, line 31: Can we use the icon for this instead? what are the cases where this is substantially different than the icon?

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -14,8 +20,17 @@ class PointDocument {
   public String poiIconColor;
   public String poiSource;
   public String poiDifficulty;
-  /** Length is in meters */
-  public double poiLength = 0;
+  public double poiLength = 0; // meters
   public String website;
   public double[] location;
+
+  // Computed ranking signals (null => omitted from JSON, ES uses missing:1.0). The poi* prefix marks
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -1,10 +1,16 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
 class PointDocument {
   public Map<String, String> name = new HashMap<String, String>();
+  // Kept separate from name and searched at a lower boost; folding variants into name hurts ranking.
```
HarelM:
```suggestion
```

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -14,8 +19,16 @@ class PointDocument {
   public String poiIconColor;
   public String poiSource;
   public String poiDifficulty;
-  /** Length is in meters */
-  public double poiLength = 0;
+  public double poiLength = 0; // meters
```
HarelM: This comment can't be seen by the IDE as opposed to the previous comment.
Please make sure not to do these kind of changes.

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -14,8 +19,13 @@ class PointDocument {
...
   public String website;
   public double[] location;
-}
\ No newline at end of file
+  public Float poiProminence;
+  public Float poiAreaNorm;
```
HarelM, line 27: In what units this field is? no units - 0 to 1?
Can we call it `poiAreaNormalized`,

src/main/java/il/org/osm/israelhiking/MainClass.java
```diff
@@ -24,9 +27,19 @@ public static void main(String[] args) throws Exception {
...
-        // Planetiler is a convenience wrapper around the lower-level API for the most
-        // common use-cases.
-        // See ToiletsOverlayLowLevelApi for an example using the lower-level API
+        // ES ingest is independent of tile zoom, so collapsing tiles to z0 builds the search index
+        // identically while making the .pmtiles archive a near-instant stub — only map tiles degrade.
+        boolean skipTiles = args.getBoolean("skip-tiles",
+                "Collapse tile output to z0 (near-instant archive) to speed up an ES-only reindex; "
+                        + "map tiles become a stub. Default false — leave off when tiles are needed.",
+                false);
+        if (skipTiles) {
```
HarelM: planetiler reports all the flags it gets, this is redundant.

src/main/java/il/org/osm/israelhiking/MainClass.java
```diff
@@ -24,9 +27,19 @@ public static void main(String[] args) throws Exception {
     }
 
     static void run(Arguments args) throws Exception {
-        // Planetiler is a convenience wrapper around the lower-level API for the most
-        // common use-cases.
-        // See ToiletsOverlayLowLevelApi for an example using the lower-level API
+        // ES ingest is independent of tile zoom, so collapsing tiles to z0 builds the search index
+        // identically while making the .pmtiles archive a near-instant stub — only map tiles degrade.
+        boolean skipTiles = args.getBoolean("skip-tiles",
```
HarelM: If you plan on keeping this, please add this to the readme.

src/main/java/il/org/osm/israelhiking/MainClass.java
```diff
@@ -38,25 +51,79 @@ static void run(Arguments args) throws Exception {
                     "bbox");
             var supportedLanguages = args.getString("languages", "Languages to support", "en,he,ru,ar,es").split(",");
             var externalFilePath = args.getString("external-file-path", "External file path", "");
+            var qrankPath = args.getString("qrank-path",
```
HarelM: Add instructions on how to use this in the readme.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
HarelM: This won't work, because I need to see the logic for both icon/color/category, feature class and ranking has similar behavior, if I can't see it in one place I loose track of what is covered by every flow.
I don't think the cache is the interesting or long running part for every feature, but I don't mind micro optimizing it. We need to make sure this doesn't create memory and GC overhead though.

src/main/java/il/org/osm/israelhiking/MainClass.java
HarelM: I've removed this flag as I'm not sure what it does to the other arguments when it is applied.
If you wish to return it make sure that only the relevant arguments are updated instead of overriding the entire args variable.

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -126,7 +214,25 @@ private static String getTargetIndexName(String indexAlias, ElasticsearchClient
...
+  /**
+   * Restore normal search-time settings before an index goes live: 
+   * re-enable periodic refresh and add one replica (createPointsIndex/createBBoxIndex disable both for faster bulk writes). 
+   */
+  public static void restoreSearchSettings(ElasticsearchClient esClient, String targetIndex)
+      throws Exception {
+    esClient.indices().putSettings(p -> p
+        .index(targetIndex)
+        .settings(s -> s
+            .refreshInterval(t -> t.time("1s"))
```
HarelM, line 235: These index is never updated, they only get rebuild right now.
Does it make sense to remove refresh interval from these indexes completely?

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -43,38 +62,25 @@ public class PlanetSearchProfile implements Profile {
...
   public PlanetSearchProfile(PlanetilerConfig config, ElasticsearchClient esClient, String pointsIndexName,
-      String bboxIndexName, String[] supportedLnaguages) {
+      String bboxIndexName, String[] supportedLnaguages, QRankIndex qrankIndex) {
     this.config = config;
     this.esClient = esClient;
     this.pointsIndexName = pointsIndexName;
     this.supportedLanguages = supportedLnaguages;
     this.bboxIndexName = bboxIndexName;
+    this.qrankIndex = qrankIndex;
+    this.bulkListener = new AccountingBulkListener(esClient, stats);
```
HarelM, line 63: bulkListener and bulkIngester should be injected to the profile.
When the planetiler finishes the run they should be invoked from outside the profile, preferable by the elastic search helper, along side all the alias and switching logic.
There's a bulk of logic at the end of the run, that shouldn't be in main class.
The previous logic was pretty slim and straight forward.
Now that it is "more complicated" it should be moved to ES helper, including waiting for the remaining requests.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -370,7 +451,7 @@ private boolean processMtbNameFeature(SourceFeature feature, FeatureCollector fe
         var minIdFeature = mergedFeature.representingFeature;
 
         var pointDocument = new PointDocument();
-        convertTagsToDocument(pointDocument, feature);
+        convertTagsToDocument(pointDocument, minIdFeature);
```
HarelM, line 454: This is a bug fix unrelated to the changes here, right?
It looks correct.

HarelM:
This is now becoming too huge of a PR to review.
I'm getting tired mid way of reviewing this.
Let's split this PR into small increments.
Each increment should be a non breaking change in the behavior of the current code.
For example:
- Fix missing indices
- Add analizers
- Fix minId bug
- Use osm cache
- Add logging support for number of indexes written
- Add area norm calcualation
- Add QRank support
- etc.

This way we can focus on each feature/fix/addition separately and I'll be able to review this thoroughly.
Sorry for the extra work, but I'm spending a lot of time re-reading things I already read.

I would focus on bug fixing for the first PRs and only afterwards add features.

HarelM:
I've updated the code to fix some things you found here.
I've added init and finalize code blocks.
I've added bridge support which is not purely related to this PR, but I wanted it added sooner rather than later.
As I said, you'll need to split your PR into smaller ones.
You can see the side of a PR by looking at the recent commits I did to main - every commit is small and focused. In order to merge your changes I'll need a similar solution.
Moreover, for things related to tags that were not supported either you or I need to update the Site repo to match the relevant tags, so make sure to split this up and add a note in the PR, that this PR adds new OSM tag support. An example I saw in the code is `volcano` which is similar to peak. Which I believe to be a good addition to the search results, but we need to make sure Site is updated with this information as well.

sergeytkachenko:
Splitting this into small, independent, non-breaking PRs as requested, bugfixes first. Closing this one in favor of:

- #68 - Make indexing lossless: bulk ingest with surfaced failures (bugfix)
- #69 - Treat natural=volcano like natural=peak in the icon classifier (bugfix)
- #70 - Add --skip-tiles to collapse tiles for a faster ES-only reindex
- #71 - Add index-time ranking signals (prominence, QRank, feature_class, area, population, alt_names)

Each PR is a single commit, merges cleanly into main on its own, and carries its own tests. The review comments from this PR are addressed in the corresponding split PR (each description has an "addressed review notes" section). #71 is the one with the intended Hebrew behavior change and the new OSM tag support that the Site repo needs to consume.

PR #68, opened 2026-06-20
Make indexing lossless: bulk ingest with surfaced failures

sergeytkachenko:
## What

The indexer fire-and-forgot each `esClient.index()` call and swallowed every error, so any document that failed to index silently disappeared from search. This is the root cause behind missing results.

## Change

Point and bbox insertion now go through a `BulkIngester` whose listener (`AccountingBulkListener`) counts every outcome and surfaces failures instead of swallowing them:

- Transient transport errors (429/5xx, timeouts) are retried with exponential backoff and jitter; per-item 4xx (mapping/parse) are charged as genuine failures and never retried.
- The ingester, listener and counters are owned by the run context in `ElasticsearchHelper` (built in `initRun`, drained and closed in `finalizeRun` before the alias swap). The profile only adds documents to the injected ingester.
- `finalizeRun` logs one summary line (emitted/indexed/failed). There is no promotion gate; a mismatch is reported in that line, not fail-closed.
- `abortRun` releases the ingester and retry threads if the build throws, so they cannot leak (no alias is swapped in that case).

## Tests

`BulkIngesterAccountingTest` and `IndexerResilienceTest` are pure unit tests (Mockito, no live cluster). The retry state machine is exercised through the real `afterBulk(...)` path with a mocked client, so there is no test-only seam in production code. `mvn test` passes.

Non-breaking and independent of the other split PRs.

## Review notes addressed (from #63)

- Inject the bulk listener/ingester and invoke them from the ES helper, not the profile or `MainClass`: done; `MainClass` is thin again.
- Drop the gate, the delete/update counts and the index-OK diagnostics (`reconcileLivePoints`, `getLiveAliasDocsStats`, `assertSafeToReindex`): all removed; a single end-of-run stats line replaces them.
- Remove the one-line failure-counter method: inlined at the call sites.
- Nothing should exist solely for tests: removed the test-only `resubmitWithBackoff(..., BulkAttempt, BackoffSleep)` overload and its interfaces; the retry path is tested through `afterBulk`.
- Comments too verbose, no ADR references, no HTML: comment-to-code ratio trimmed; rationale moved to method javadoc.

Part of splitting #63 into small, independent, non-breaking increments (bugfixes first). Companion PRs: volcano icon, `--skip-tiles`, ranking signals.

### DIFF
--- a/src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
+++ b/src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
@@ -9,11 +9,15 @@ import org.apache.http.HttpHost;
 import org.elasticsearch.client.RestClient;
 
 import co.elastic.clients.elasticsearch.ElasticsearchClient;
+import co.elastic.clients.elasticsearch._helpers.bulk.BulkIngester;
 import co.elastic.clients.json.jackson.JacksonJsonpMapper;
+import co.elastic.clients.transport.BackoffPolicy;
 import co.elastic.clients.transport.ElasticsearchTransport;
 import co.elastic.clients.transport.rest_client.RestClientTransport;
 
 public class ElasticsearchHelper {
+
+  private static final Logger LOGGER = Logger.getLogger(ElasticsearchHelper.class.getName());
 
   public static record ElasticRunContext(
       ElasticsearchClient esClient,
@@ -21,7 +25,10 @@ public class ElasticsearchHelper {
       String bboxIndexAlias,
       String pointsIndexTarget,
       String bboxIndexTarget,
-      String[] supportedLanguages) {
+      String[] supportedLanguages,
+      BulkIngester<Void> bulkIngester,
+      AccountingBulkListener bulkListener,
+      IndexingStats stats) {
   }
 
   /**
@@ -152,12 +159,40 @@ public class ElasticsearchHelper {
     var targetPointsIndex = ElasticsearchHelper.createPointsIndex(esClient, pointsIndexAlias,
         supportedLanguages);
     var targetBBoxIndex = ElasticsearchHelper.createBBoxIndex(esClient, bboxIndexAlias, supportedLanguages);
+    var stats = new IndexingStats();
+    var bulkListener = new AccountingBulkListener(esClient, stats);
+    BulkIngester<Void> bulkIngester = BulkIngester.of(b -> b
+        .client(esClient)
+        .maxOperations(5_000)
+        .maxSize(5 * 1024 * 1024)
+        .maxConcurrentRequests(4)
+        // noBackoff pinned: the listener owns retry, so the ingester must not double-retry.
+        .backoffPolicy(BackoffPolicy.noBackoff())
+        .listener(bulkListener));
     return new ElasticRunContext(esClient, pointsIndexAlias, bboxIndexAlias, targetPointsIndex, targetBBoxIndex,
-        supportedLanguages);
+        supportedLanguages, bulkIngester, bulkListener, stats);
   }
 
   public static void finalizeRun(ElasticRunContext context) throws Exception {
+    context.bulkIngester().flush();
+    context.bulkIngester().close();
+    context.bulkListener().awaitRetries();
+
+    var stats = context.stats();
+    LOGGER.info(() -> "Indexing finished: emitted=" + stats.getEmittedCount()
+        + " indexed=" + stats.getIndexedCount() + " failed=" + stats.getFailedCount());
+
     ElasticsearchHelper.switchAlias(context.esClient, context.pointsIndexAlias(), context.pointsIndexTarget());
     ElasticsearchHelper.switchAlias(context.esClient, context.bboxIndexAlias(), context.bboxIndexTarget());
+  }
+
+  // Releases ingester and retry executor without swapping aliases; safe in a finally after a failed run.
+  public static void abortRun(ElasticRunContext context) {
+    try {
+      context.bulkIngester().close();
+    } catch (Exception e) {
+      LOGGER.warning("Bulk ingester close failed during abort: " + e.getMessage());
+    }
+    context.bulkListener().awaitRetries();
   }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/il/org/osm/israelhiking/MainClass.java
```diff
@@ -52,9 +52,17 @@ static void run(Arguments args) throws Exception {
                 planetiler.addGeoJsonSource("external", Path.of(externalFilePath));
             }
             planetiler.overwriteOutput(Path.of("data", "target", PlanetSearchProfile.POINTS_LAYER_NAME + ".pmtiles"));
-            planetiler.run();
 
-            ElasticsearchHelper.finalizeRun(context);
+            boolean finalized = false;
```
HarelM: Move this logic into the finalizeRun method, there should not be a try-catch with-in a try-catch here.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -657,13 +671,17 @@ private void insertBboxToElasticsearch(SourceFeature feature, String[] supported
...
-          .id(sourceFeatureToDocumentId(feature))
-          .document(bbox));
+      bboxDocId = sourceFeatureToDocumentId(feature);
     } catch (Exception e) {
-      // swallow
+      LOGGER.warning(() -> "Failed to build bbox document: " + e.getMessage());
+      return;
     }
+    final BBoxDocument document = bbox;
+    addToIngester(BulkOperation.of(op -> op
```
HarelM: Why is this part out of the try-catch? I believe the main reason for the try-catch is the document indexing. If this is no longer needed, please remove the try-catch.

HarelM:
Please checkout the coverage report and see if there's room for improvement. Make sure that the catches that are not covered are needed and it is not just defensive programming.
Also please make the log messages a bit shorter, it looks like they take a lot of the code and I'm not sure who will read them as we are not monitoring this process closely.

src/main/java/il/org/osm/israelhiking/MainClass.java
sergeytkachenko: Done. `MainClass.run` is now two stacked try-with-resources blocks (esClient → run context) with no `catch` — the nested try-catch is gone and the finalize/abort logic lives behind the context lifecycle: `ElasticRunContext implements AutoCloseable`, so `finalizeRun(context)` runs on the happy path and `close()` aborts (no alias swap) on any throw. `MainClass` is thin again. (commit ba265e4)

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Done. `insertBboxToElasticsearch` is now a single `try` that wraps the whole bbox build **and** the `addToIngester` enqueue — the enqueue is no longer outside it. Since indexing no longer throws (the ingester swallows + counts internally), the only reason the `try` remains is to keep the lossless guarantee: any failure during the build (incl. `GeometryException`) is charged to `failedCount` and logged, instead of a silent `return`. So the try is kept deliberately, but tightened to one block with the enqueue inside. (commit ba265e4)

sergeytkachenko:
Pushed follow-ups addressing the remaining review points plus an internal code-review pass.

**Your summary comments**
- **Coverage / are the uncovered catches needed** — added `DefensiveCatchCoverageTest` that forces the two defensive catches with a genuine trigger: the `addToIngester` enqueue-failure path (mocked ingester throws on `add`) and the `ElasticRunContext.close()` ingester-close-failure path. Both are real (closed/failed ingester during the run, and abort not masking the build error), not dead defensive code.
- **Shorter log messages** — the end-of-run summary is one line (`emitted= indexed= failed=`); per-event listener logs trimmed. I also removed the explanatory in-code comments this PR had added (commit e196597), per your "comments too verbose" note from #63.

**Internal code-review fixes (commits 9861ea7, e196597)**
- **Double close on the happy path** — `finalizeRun` closed the ingester + drained, then try-with-resources `close()` did it again, logging a misleading `"close failed during abort"` WARNING on every successful build. Now a one-shot `tryClaimIngesterClose()` + a `drained` guard make both idempotent; pinned by a new test.
- **Listener logger** — `AccountingBulkListener` was logging under the `PlanetSearchProfile` logger; fixed to its own class so per-class log routing works.
- **Null-safe bulk items** — `afterBulk`/`resubmitWithBackoff` now normalize `response.items()` to an empty list, avoiding an NPE on the ingester thread for a malformed response.

**Left for your call (not changed):**
- Retryable status set is `429/5xx`; I did *not* add `408`/`500` — that's a retry-policy decision, happy to add if you want transient 500s (circuit breaker) retried.
- No promotion gate before the alias swap on a drain-timeout — kept as the documented "report, don't fail-closed" behavior.

`mvn test` green (29 unit tests in the affected suites).

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -9,19 +9,38 @@
...
       ElasticsearchClient esClient,
       String pointsIndexAlias,
       String bboxIndexAlias,
       String pointsIndexTarget,
       String bboxIndexTarget,
-      String[] supportedLanguages) {
+      String[] supportedLanguages,
+      BulkIngester<Void> bulkIngester,
+      AccountingBulkListener bulkListener,
+      IndexingStats stats) implements AutoCloseable {
```
HarelM: This is a record, I don't think it should autoclosabele.
But I think it might make sense to have `bulkListener` autocolosable.

sergeytkachenko:
Done — addressed your last comment. `ElasticRunContext` is no longer `AutoCloseable`; the close/drain lifecycle now lives in `AccountingBulkListener`, which implements `AutoCloseable`. The listener owns the ingester (`attachIngester`, `volatile`) and `MainClass` drives it via try-with-resources; `finalizeRun` just calls `bulkListener.close()`. `close()` is idempotent so the finalize and abort paths can both call it, and the tests assert it drains retries and closes the ingester exactly once.

Squashed the review follow-ups into a single commit. CI is green (Java + Docker), coverage checks pass. Ready for another look whenever you have a moment.

HarelM (approved):

HarelM:
Can you check if there's coverage improvement that can be done? It looks like a lot of tests but coverage is at 80-90, so I would like to have less tests and more coverage if possible.

PR #71, opened 2026-06-20
Add index-time ranking signals (prominence, QRank, feature_class, area, population, alt_names)

sergeytkachenko:
## What

Adds the build-time signals the query-side ranking reads. Every field is additive: an absent field falls back to Elasticsearch `missing:1.0`, so existing queries keep working.

- `OsmTagUtils`: pure tag helpers - `parseFirstNumber`, `classifyFeatureClass` (coarse feature type), `buildAltNames` (variant names), `computePopulation` (settlement ladder with an overflow guard).
- `PointDocument` and mapping: `poiProminence`, `poiAreaNormalized`, `intermittent`, `population`, `poiFeatureClass`, `alt_names`; `@JsonInclude(NON_NULL)`.
- Analyzers: an edge-ngram `name.<lang>.prefix` subfield for as-you-type recall, and a Hebrew doubled-matres fold on `name.he`.
- `ProminenceCalculator` and `QRankIndex`: a composite [0,1] prominence from a feature-class prior, elevation, Wikimedia QRank and metadata richness, floored so a query-time multiply never zeroes a hit. The prior keys off `poiFeatureClass` (the single `classifyFeatureClass` mapping), not raw tags. QRank loads from an optional `--qrank-path` (empty means lookups return 0).
- Every emit path, including the non-icon path, routes through `convertTagsToDocument`, so all features carry the signals.

## One intended behavior change: Hebrew

`name.he` moves to a `hebrew_analyzer` that folds doubled matres (a doubled vav to a single vav, a doubled yod to a single yod; single letters are never dropped). Hebrew searches now match spelling variants they previously missed. This pairs with the Site `query_norm`; both halves must deploy together.

## New OSM tag support

The classifier recognizes tag values the index did not surface before (volcano, canyon, plateau, arch, cave, wetland and more), `boundary=national_park|protected_area`, and variant-name tags. The Site repo must be updated to consume the new fields and values for ranking.

## Tests

`ParseFirstNumber`, `SetFeatureClass`, `AddAltNames`, `ComputePopulation`, `HebrewMatresRule`, `PointsIndexMapping`, `PointsAnalyzerMapping`, `ProminenceCalculator`, `InsertProminenceFloor`, `QRankIndex`, `NormalizeArea`. All 125 unit tests pass.

## Review notes addressed (from #63)

- Prefix computed (non-OSM) fields with `poi`, and use `poiFeatureClass` rather than `feature_class`: done.
- Rename `poiAreaNorm` to `poiAreaNormalized`: done.
- Simplify the QRank parsing (use split, not manual indexing): `load` parses the two-column CSV with `split`, and `getByWikidata` splits a multi-value tag and streams to the max rank.
- Explain the 32M capacity hint: it is a hint to avoid rehashing during load; the map grows for a larger file.
- Is wikidata counted twice: documented; wikidata as a richness flag and wikidata as the QRank key are intentionally separate.
- Remove `baseScore` and keep icon/feature_class/ranking together: the prominence prior now keys off `poiFeatureClass` instead of re-deriving from raw tags, so the three flows share one source of truth.
- Add `--qrank-path` usage to the README: done, with a download and run example.
- Comments too verbose, no HTML, no ADR references: aggressive comment sweep across the new files.

Part of splitting #63 into small, independent, non-breaking increments. Companion PRs: indexer resilience, volcano icon, `--skip-tiles`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Additional arguments to this wrapper besides the Planetiler's arguments:
 | `es-points-index-alias` | The alias of the index to insert points into, it will create "1" and "2" suffix for the relevant index before switching | `points` |
 | `es-bbox-index-alias` | The alias of the index to insert bounding boxes into, it will create "1" and "2" suffix for the relevant index before switching | `bbox` |
 | `external-file-path` | External geojson file path to allow adding non OSM features to the search and POIs. these features should have a specific format | "empty" |
+| `qrank-path` | Path to a `qrank.csv.gz` file (Wikimedia pageview ranks, from https://qrank.toolforge.org) used to boost the prominence of well-known features. Leave empty to build without QRank — all QRank lookups then contribute 0. | "empty" |
 
 To run using docker (While having a local elasticsearch running at 9200):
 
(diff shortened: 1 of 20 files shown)


HarelM:
This PR is still too big. Please split this up more.

HarelM (changes requested):
PR is still too big.

sergeytkachenko:
Split into three small, independent, non-breaking PRs as requested, each comment-free and rebased on current `main`:

- #72 — OSM tag enrichment fields (feature_class, alt_names, area, population)
- #73 — prefix + Hebrew-matres analyzers (stacked on #72)
- #74 — prominence ranking signal with optional QRank (stacked on #73)

Review order: #72 → #73 → #74. A self-review also caught and fixed a QRank overflow (top-ranked entities were being dropped) and gave `alt_names.he` the same Hebrew analyzer as `name.he`.

Closing this in favor of the three above.

PR #72, opened 2026-06-25
Add OSM tag enrichment fields (feature_class, alt_names, area, population)

sergeytkachenko:
## What

First of three small, non-breaking PRs that replace #71 (which was too large to review). Adds the build-time OSM tag enrichment fields the query-side ranking reads.

Every field is additive: an absent field falls back to Elasticsearch `missing:1.0`, so existing queries and old documents keep working unchanged.

- **`OsmTagUtils`** — pure tag helpers: `parseFirstNumber` (free-text number parsing), `classifyFeatureClass` (coarse feature type incl. volcano/canyon/plateau/arch/cave/wetland and `boundary=national_park|protected_area`), `buildAltNames` (variant names), `computePopulation` (settlement ladder with an overflow guard).
- **`PointDocument` + mapping** — new nullable fields `poiFeatureClass`, `poiAreaNormalized`, `intermittent`, `population`, `alt_names`; `@JsonInclude(NON_NULL)` so a null field is omitted and ES uses `missing:1.0`.
- **`PlanetSearchProfile`** — enrichment wiring; every emit path (including the non-icon path) routes through `convertTagsToDocument`, so all features carry the signals.

## Tests

`ParseFirstNumber`, `SetFeatureClass`, `AddAltNames`, `ComputePopulation`, `PointsIndexMapping`. All green.

## Stack

This is the base of a 3-PR stack that supersedes #71:
1. **this PR** — enrichment fields
2. analyzers (prefix + Hebrew matres) — stacked on this
3. prominence + QRank — stacked on the analyzers PR

No code comments beyond one-line field/units notes, per review feedback on #63/#71.

PR #73, opened 2026-06-25
Add prefix and Hebrew-matres analyzers for the points index

sergeytkachenko:
> 🚧 **Draft until #72 merges.** This branch is stacked on #72, so the diff above currently includes #72's files too. Once #72 is merged, I'll rebase this onto `main` (its diff collapses to just the analyzer files) and mark it ready. Review **#72 first**.

---

## What

Second of three small PRs replacing #71. Adds the points-index analyzers.

- An edge-ngram `name.<lang>.prefix` subfield for as-you-type recall.
- A Hebrew **doubled-matres** fold on `name.he` (doubled vav → single vav, doubled yod → single yod; single letters are never dropped), so Hebrew searches match spelling variants (כתיב מלא / חסר) they previously missed.
- The same Hebrew analyzer + `.prefix` treatment is mirrored onto `alt_names.<lang>`.

Additive: unmapped/old documents are unaffected.

## ⚠️ One intended behavior change — Hebrew

`name.he` (and `alt_names.he`) move to a `hebrew_analyzer`. This pairs with the Site-side `query_norm`; **both halves must deploy together**. Hebrew results intentionally change for the better; the new exact-name bonus is additive ≥ 0, so this is not a regression for old clients.

## Stack — review #72 first

> **Note on the diff:** this branch is stacked on **#72** (enrichment fields). GitHub shows the cumulative diff because cross-fork PRs can't target a branch that isn't in this repo yet. **This PR's own change is only the analyzer additions:**
> - `ElasticsearchHelper.java` — analyzer/normalizer/char-filter definitions, the `name.<lang>.prefix` subfield, and the `alt_names.<lang>` Hebrew mirror
> - `HebrewMatresRuleTest.java`, `PointsAnalyzerMappingTest.java`
>
> Everything else in the diff belongs to #72. Once #72 merges, this PR's diff collapses to just the above. Merge order: #72 → this → prominence.

No code comments beyond one-line WHY notes, per review feedback on #63/#71.

### DIFF
--- a/src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
+++ b/src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
@@ -15,6 +15,22 @@ import co.elastic.clients.transport.rest_client.RestClientTransport;
 
 public class ElasticsearchHelper {
 
+  // Doubled-only: folding a single vav/yod would merge real homographs. "\\u" reaches ES as the literal Lucene escape.
+  static final String HEBREW_VAV = "ו";
+  static final String HEBREW_YOD = "י";
+  static final String HEBREW_DOUBLED_VAV_PATTERN = "\\u05D5\\u05D5";
+  static final String HEBREW_DOUBLED_YOD_PATTERN = "\\u05D9\\u05D9";
+
+  /** Reference implementation of the doubled-only matres fold, so a unit test can assert it without a live ES. */
+  static String applyHebrewMatresDoubledOnly(String input) {
+    if (input == null) {
+      return null;
+    }
+    return input
+        .replaceAll("וו", HEBREW_VAV)
+        .replaceAll("יי", HEBREW_YOD);
+  }
+
   public static record ElasticRunContext(
       ElasticsearchClient esClient,
       String pointsIndexAlias,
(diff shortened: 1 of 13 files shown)


sergeytkachenko:
Converted to **draft** to avoid review noise.

This PR is stacked on #72, and since I don't have push access to this repo I can't target its branch directly — so GitHub shows the cumulative diff (all of #72's files appear here too). Its **own** change is only:
- `ElasticsearchHelper.java` — analyzer/normalizer/char-filter definitions, `name.<lang>.prefix`, and the `alt_names.<lang>` Hebrew mirror
- `HebrewMatresRuleTest.java`, `PointsAnalyzerMappingTest.java`

Once #72 merges I'll rebase this onto `main` (the diff collapses to just the above) and mark it ready. **Please review #72 first.**

src/main/java/il/org/osm/israelhiking/ElasticsearchHelper.java
```diff
@@ -53,25 +69,87 @@ public static String createPointsIndex(ElasticsearchClient esClient, String inde
                         .patternReplace(pr -> pr
                             .pattern("[\\u05B0-\\u05C7]")
                             .replacement(""))))
+                .charFilter("hebrew_matres", cf -> cf
```
HarelM, line 72: Should be renamed to hebrew_matters_vav in all places in this file.

sergeytkachenko:
Closing in favor of #75 — you migrated the analyzers, mapping and document fields from this branch into #75 (and renamed the matres char-filter to `hebrew_matres_vav`, which also fixes a naming mismatch this branch had). So everything here is now covered there and keeping #73 open would only create a conflicting duplicate.

One thing #75 dropped that was here: the analyzer unit coverage (`PointsAnalyzerMappingTest`, `HebrewMatresRuleTest`). Happy to re-add a mapping/analyzer test against the Testcontainers ES in a follow-up so the Hebrew matres fold and the `.prefix` / `alt_names` mappings stay pinned — without the Java reference impl you (rightly) didn't want to keep around just for a test. Let me know if you'd like that.

PR #75, opened 2026-06-26
Normalizers, area and intermittent

HarelM:
- Replaces #73


HarelM:
@sergeytkachenko I've took some of your code as part of PR #73 and migrated it here.
This adds the relevant indexing mapping for elasticsearch.
It adds the properties in the document.
Fills up area and intermittent.
Note that I have placed the area normalized only on features that I think the area is relevant for (non linear features and points).
Let me know if there's something here I need to change.
I'm not sure there's a need for the defensive programing around normalized area NaN as according to the tests here this is dead code.

PR #77, opened 2026-06-28
Manual search-relevance check (test-only)

sergeytkachenko:
## Summary

Dramatically reduced per @HarelM's review. Replaces the multi-class relevance
runner + HTML report with a single JUnit `@TestFactory`: a flat list of cases
(`searchTerm`, `center`, `zoom`, `uiLanguage`, `expectedTarget`, `radiusMeters`,
`topN`); for each, call the live `/api/search`, take the top-N hits, and assert
one lands within `radiusMeters` of the expected target (haversine). The failure
message reports the closest hit's title for inspection.

- **Test-only** — no new production classes (was 10), so coverage is unaffected.
- Tagged `relevance`, excluded from the default `unit` / `e2e` groups; runs only
  on demand via `mvn -Prelevance -Drelevance.endpoint=…` or the `workflow_dispatch`
  action. It does not gate normal CI.
- ~555 lines across 4 files (was ~2878 across 19).

## Why proximity, not a name match

The assertion is on coordinate proximity, not the hit's name: the gold set is
multilingual and includes nicknames, abbreviations and typos ("Tour Eiffel",
"RMNP", "IBT", "Humphrys Peak") where a title match would false-fail.

## Note on the live API

Per the swagger, `GET /api/Search/{term}` accepts only `term` + `language` and
ranks purely by name-match score — there is no location-bias parameter, so the
`center`/`zoom` fields are sent but currently ignored by the API. The check
therefore asks "is the expected feature the top global name match", and many
generic-name rows fail until ranking/indexing improvements land. Run it manually
to read the current relevance score.


HarelM:
Can this be dramatically reduced please?
I generally want something that should be fairly simple:
a list of something like:
```
searchTerm: "xxx",
center: [1,2],
zoom: 3,
uiLanguage: "en"
expectedTarget: [5,6]
```

For each item in the list, do the search against the relevant URL and return true/false if one of the first results are matching/closed-to the target.

Is it not what the current test is testing?

sergeytkachenko:
@HarelM 
Done — reduced to a single test in exactly that shape; please take another look.

`SearchRelevanceTest` reads a flat list (`searchTerm`, `center`, `zoom`, `uiLanguage`, `expectedTarget`, `radiusMeters`, `topN`) and for each one hits the live `/api/search`, passing when a top-N result is within `radiusMeters` of `expectedTarget`. Test-only, ~545 lines across 4 files (was ~2878 / 19), tagged `relevance` so it stays out of normal CI.

Re "is it not what the current test is testing?" — no, `E2ETest` only runs the indexing build (`MainClass.main(--download …)`) and asserts nothing about search results; this is the first test that scores relevance against expected targets.

HarelM (approved):
This is great! Thanks!

PR #78, opened 2026-06-28
Index-time prominence + QRank ranking signals (poiProminence) + population

sergeytkachenko:
> **Stacked on #79.** The indexer-resilience / lossless-reindex half of the original PR moved to
> #79 (review that first). The 3 resilience commits still show in this PR's diff until #79 merges;
> once it does, this branch is rebased onto `main` to drop them, leaving only the 5 ranking-signal
> commits below.

## What & why

The **index-side companion** for the query-side relevance work in **IsraelHikingMap/Site#2543**.
That PR ranks search results with a `function_score` that reads a build-time **`poiProminence`**
field; this PR is what actually **computes and populates** that field (plus `population`).

Builds on the renamed ranking-signal scheme already merged in `#75` (`poiProminence`,
`poiFeatureClass`, `population`, `poiAreaNormalized`, `intermittent`): `#75` declared the fields,
this PR fills in the prominence/QRank computation and the ES mapping for `poiProminence`. The
indexer-resilience guarantees this work relies on (lossless reindex) are in **#79**.

## What's in it

**Build-time prominence (`poiProminence`)** — `ProminenceCalculator`:

```
prom = clamp01(0.05 + 0.45*base + 0.40*qnorm + 0.10*meta)   // floor 0.05
```
- `base` from `natural`/`place`/`boundary`/`tourism`/`historic`/`waterway` + normalized elevation
  (peaks scale with height; city/town/village tiers; parks; viewpoints; springs; …).
- `qnorm` from the QRank signal, gated on `wikidata` (`log(qrank)/log(3e6)`).
- `meta` from `image` / `website` / `wikidata` presence.
- **Insert-path floor** so every emitted doc carries a `poiProminence > 0` — a query-time
  `field_value_factor` multiply can never zero a result out.

**QRank signal** — `QRankIndex` + new `--qrank-path` arg (documented in the README). Optional:
when no file is passed it runs with an **empty index** (lookups return 0), so the build never
depends on the QRank file being present.

**`population`** — populated for place/admin features.

## Tests

4 new `@Tag("unit")` classes specific to this PR, all green in the default Surefire group (zero
network): `ProminenceCalculatorTest`, `QRankIndexTest`, `ParseFirstNumberTest`,
`InsertProminenceFloorTest`. (The BulkIngester accounting tests live in #79.) Full suite:
**54 tests, 0 failures**.

## Non-breaking

- `poiProminence` is `@JsonInclude(NON_NULL)` and floored on the insert path; the query side
  treats a missing value as `1.0`, so old docs degrade gracefully until the next reindex.
- ES mapping is additive: `poiProminence` searchable; raw components stored `index:false` so
  weights can be re-tuned without a reindex.
- The new fields do nothing until both a reindex runs **and** the query side (#2543) is deployed.

## Merge order

1. **#79** (lossless reindex) — independent indexer hardening; safe on today's index, merge first.
2. Query-side **Site#2543** — reversible binary redeploy, safe on today's index.
3. **This PR** — rebased onto a `main` that already has #79; the multi-hour reindex whose fields
   light up once #2543 is live.

## Follow-up (not in this PR)

- `country_code` — read by #2543's scoring but not yet produced here; separate enrichment.

---

> Draft — companion to IsraelHikingMap/Site#2543.

### DIFF
--- a/src/main/java/il/org/osm/israelhiking/PointDocument.java
+++ b/src/main/java/il/org/osm/israelhiking/PointDocument.java
@@ -24,7 +24,6 @@ class PointDocument {
   public String website;
   public double[] location;
   public Float poiProminence;
-  // Raw prominence components, stored index:false so the weights can be re-tuned without a reindex.
   public Float poiPromBase;
   public Float poiPromQrankNorm;
   public Float poiPromMeta;
--- a/src/main/java/il/org/osm/israelhiking/PointDocument.java
+++ b/src/main/java/il/org/osm/israelhiking/PointDocument.java
@@ -1,9 +1,12 @@
 package il.org.osm.israelhiking;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
 class PointDocument {
   public Map<String, String> name = new HashMap<String, String>();
   public Map<String, List<String>> alt_names;
@@ -21,6 +24,12 @@ class PointDocument {
   public String website;
   public double[] location;
   public Float poiProminence;
+  // Raw prominence components, stored index:false so the weights can be re-tuned without a reindex.
+  public Float poiPromBase;
+  public Float poiPromQrankNorm;
+  public Float poiPromMeta;
+  public Float poiEleNorm;
+  public Long poiQrankRaw;
   public Float poiAreaNormalized;
   public Boolean intermittent;
   public Integer population;
--- a/src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
+++ b/src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
@@ -1,37 +1,13 @@
 package il.org.osm.israelhiking;
 
-/**
- * Computes a composite {@code poiProminence} score in [0,1] for a feature, used at query time as a
- * {@code field_value_factor} multiplier so a major peak / well-known place outranks an obscure node
- * with the same name.
- *
- * <p>Formula (research-derived):
- * <pre>
- *   poiProminence = clamp01( 0.05 + 0.45*base + 0.40*qnorm + 0.10*meta )
- * </pre>
- * <ul>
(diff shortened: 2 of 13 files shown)


src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,80 @@
...
+
+final class ProminenceCalculator {
+
+  static final double QRANK_REF = 3_000_000.0;
+  static final double MAX_ELE_M = 8849.0;
+  static final double FLOOR = 0.05;
+
+  private ProminenceCalculator() {}
+
+  static final class Result {
```
HarelM, line 11: `record`?

src/main/java/il/org/osm/israelhiking/PointDocument.java
```diff
@@ -21,6 +24,11 @@ class PointDocument {
   public String website;
   public double[] location;
   public Float poiProminence;
+  public Float poiPromBase;
```
HarelM, line 27: Are all these fields needed? I think we agreed in a previous review that they are debug only, aren't they?

src/test/java/il/org/osm/israelhiking/BulkIngesterAccountingTest.java
HarelM, line 1: I don't think this is related to this PR, isn't it?

src/test/java/il/org/osm/israelhiking/EmittedCountInvariantTest.java
HarelM, line 1: Same for this file.

HarelM:
Looks like a mixed PR of some weird git state... Please split this PR into prominence and lossless reindex...

sergeytkachenko:
Closing in favor of #81, which re-derives just the prominence / QRank / population work cleanly on top of `main` (the fields and mappings are already merged via #72/#73/#75, so #81 only adds the computation).

This addresses the review here:
- **Split out** — the lossless-reindex / `BulkIngester` half is now tracked in #80, separate from ranking.
- **Debug-only `PointDocument` breakdown fields removed.**
- **`ProminenceCalculator.Result` is now a `record`.**

Thanks @HarelM for the review.

PR #79, opened 2026-06-28
Lossless reindex: route ES inserts through BulkIngester with fail-on-loss

sergeytkachenko:
## What & why

Split out from #78 as the first of two stacked PRs — the **indexer-resilience** half, with
no ranking-signal or scoring changes. Once this lands, #78 is rebased down to just the
`poiProminence` / QRank / `population` work.

Today every POI is indexed with a per-document `esClient().index(...)` call that **swallows
exceptions**: a failed insert is silently lost, yet the alias still swaps onto the partial
index. This PR makes a reindex **lossless or loud**, and is independently shippable — it does
not depend on the query-side relevance work (IsraelHikingMap/Site#2543).

## What's in it

**BulkIngester with lossless accounting** — replace the per-document index calls with a single
thread-safe `BulkIngester<Void>` (`maxOperations=5_000`, `maxSize≈5 MB`,
`maxConcurrentRequests=4`). A static `AccountingBulkListener` with `LongAdder` counters surfaces
and classifies every failure by destination index (per-item `WARNING`, whole-batch `SEVERE`) —
never swallowed. The invariant `emitted == indexed + failed` is observable via getters.

**Flush before the alias swap** — `ElasticsearchHelper.finalizeRun(context, profile)` drains the
ingester (`flush`/`close` blocks until in-flight requests and their listener callbacks complete),
refreshes both indices, restores search-time settings, then swaps the alias.

**Build-time index tuning** — `createPointsIndex` / `createBBoxIndex` set `refresh_interval=-1`
and `number_of_replicas=0` while building (the index isn't served until the alias swaps, so this
removes the biggest per-document bulk-indexing overhead); `restoreSearchSettings` restores
`refresh_interval=1s` / `number_of_replicas=1` before the swap.

**Fail on dropped POIs** — if any **points** document was dropped, `finalizeRun` leaves the live
points alias on the previous complete index and `MainClass` throws `IllegalStateException` with
the emitted / indexed / failed-points / failed-bbox counts, so CI fails non-zero instead of
shipping a partial points index. bbox `geo_shape` rejects are warn-but-tolerated (the bbox alias
still advances). An unknown/undeterminable destination is classified **fail-closed as POINTS**,
so a future op type can't downgrade a points loss to a tolerated bbox warning.

## Tests

2 new `@Tag("unit")` classes, zero network: `BulkIngesterAccountingTest`,
`EmittedCountInvariantTest` (drives the real `emittedCount` `LongAdder` through the production
insert paths and asserts the lossless invariant). Full suite green — `mvn test`: **17 tests, 0
failures**.

## No new dependencies

`BulkIngester` ships with the Elasticsearch Java client already on the classpath.


src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -27,12 +29,24 @@
...
+
 import il.org.osm.israelhiking.ElasticsearchHelper.ElasticRunContext;
 
 public class PlanetSearchProfile implements Profile {
+  private static final Logger LOGGER = Logger.getLogger(PlanetSearchProfile.class.getName());
+
   private PlanetilerConfig config;
   private ElasticRunContext context;
 
+  private final BulkIngester<Void> bulkIngester;
```
HarelM: This was supposed to be in a different file, it was already fixed, why are we in a previous version of the code? Should I not be reviewing this? I'm confused...

sergeytkachenko: I will check. 
This is draft, please wait ready for review.

HarelM: Sure, no problem.

sergeytkachenko:
Closing in favor of #80, which tracks the underlying problem (the per-document `index()` swallow / lossy reindex).

It's low priority — failed inserts are rare in practice — and the review here got tangled up with a stale branch state. When this is prioritized, the implementation can be revived from #68 (commit `c0ecfc5`): separate `AccountingBulkListener` / `IndexingStats`, retry + backoff, and an `AutoCloseable` drain.

Thanks @HarelM for the review.

PR #81, opened 2026-06-28
Index-time prominence + QRank ranking signal (poiProminence) + population

sergeytkachenko:
## What

Index-time **prominence** + **QRank** ranking signal (`poiProminence`) and **population** — populating fields that `main` already maps (via #72/#73/#75) but never fills.

This is a clean re-derivation of #78 on top of `main`, addressing that review:
- **Prominence only.** The lossless-reindex / `BulkIngester` work is split out and tracked in #80 — it is **not** in this PR.
- **Debug breakdown fields removed** from `PointDocument` (they were debug-only).
- **`ProminenceCalculator.Result` is a `record`.**
- No unrelated indexer tests.

## Signals

- `QRankIndex` — loads a gzipped `qrank.csv.gz` (`--qrank-path`, optional) into a primitive `long` map keyed by Wikidata QID. Omit the flag and the build runs unchanged without the file.
- `ProminenceCalculator` — base score (feature type + elevation) + QRank (only when wikidata is present) + metadata → a `0..1` `poiProminence`.
- `population` — parsed from `population` / `place` tags with per-place fallbacks.
- `@JsonInclude(NON_NULL)` on `PointDocument` + a `flooredProminence` net so a null prominence can't fall back to `field_value_factor` `missing:1.0` and outrank scored features.

No transport/indexer changes — inserts use the existing path.

## Tests

`ProminenceCalculatorTest`, `QRankIndexTest`, `ParseFirstNumberTest`, `InsertProminenceFloorTest` (`@Tag("unit")`, zero network). `mvn test`: **46 tests, 0 failures**.

Supersedes #78 (which will be closed).

### DIFF
--- a/src/main/java/il/org/osm/israelhiking/OsmFeatureClassifier.java
+++ b/src/main/java/il/org/osm/israelhiking/OsmFeatureClassifier.java
@@ -1,0 +1,276 @@
+package il.org.osm.israelhiking;
+
+import com.onthegomap.planetiler.reader.WithTags;
+
+final class OsmFeatureClassifier {
+
+  static final double DEFAULT_BASE_SCORE = 0.25;
+
+  enum Category {
+    NATURE_RESERVE("icon-leaf", "#008000", "Other", 0.80, 0.0),
+    PROTECTED_NODE("icon-leaf", "#008000", "Other", DEFAULT_BASE_SCORE, 0.0),
+    ROUTE_HIKING("icon-hike", "black", "Hiking", DEFAULT_BASE_SCORE, 0.0),
+    ROUTE_BICYCLE("icon-bike", "black", "Bicycle", DEFAULT_BASE_SCORE, 0.0),
+    ROUTE_4X4("icon-four-by-four", "black", "4x4", DEFAULT_BASE_SCORE, 0.0),
+    HISTORIC_RUINS("icon-ruins", "#666666", "Historic", 0.55, 0.0),
+    HISTORIC_ARCHAEOLOGICAL("icon-archaeological", "#666666", "Historic", 0.55, 0.0),
+    HISTORIC_MEMORIAL("icon-memorial", "#666666", "Historic", 0.55, 0.0),
+    HISTORIC_TOMB("icon-cave", "black", "Natural", 0.55, 0.0),
+    PICNIC("icon-picnic", "#734a08", "Camping", DEFAULT_BASE_SCORE, 0.0),
+    NATURAL_CAVE("icon-cave", "black", "Natural", 0.30, 0.0),
+    NATURAL_SPRING("icon-tint", "#1e80e3", "Water", 0.30, 0.0),
+    NATURAL_TREE("icon-tree", "#008000", "Natural", DEFAULT_BASE_SCORE, 0.0),
+    NATURAL_FLOWERS("icon-flowers", "#008000", "Natural", DEFAULT_BASE_SCORE, 0.0),
+    NATURAL_WATERHOLE("icon-waterhole", "#1e80e3", "Water", DEFAULT_BASE_SCORE, 0.0),
+    PEAK("icon-peak", "black", "Natural", 0.30, 0.55),
+    RIDGE("icon-peak", "black", "Natural", DEFAULT_BASE_SCORE, 0.0),
+    WATER_BODY("icon-tint", "#1e80e3", "Water", DEFAULT_BASE_SCORE, 0.0),
+    MAN_MADE_WATER_WELL("icon-water-well", "#1e80e3", "Water", DEFAULT_BASE_SCORE, 0.0),
+    MAN_MADE_CISTERN("icon-cistern", "#1e80e3", "Water", DEFAULT_BASE_SCORE, 0.0),
+    WATERFALL("icon-waterfall", "#1e80e3", "Water", 0.30, 0.0),
+    WATERWAY_RELATION("icon-river", "#1e80e3", "Water", DEFAULT_BASE_SCORE, 0.0),
+    PLACE_CITY("icon-home", "black", "Wikipedia", 1.00, 0.0),
+    PLACE_TOWN("icon-home", "black", "Wikipedia", 0.80, 0.0),
+    PLACE_VILLAGE("icon-home", "black", "Wikipedia", 0.55, 0.0),
+    PLACE_HAMLET("icon-home", "black", "Wikipedia", 0.35, 0.0),
+    PLACE_OTHER("icon-home", "black", "Wikipedia", 0.45, 0.0),
+    PLACE_BLANK("icon-home", "black", "Wikipedia", DEFAULT_BASE_SCORE, 0.0),
+    VIEWPOINT("icon-viewpoint", "#008000", "Viewpoint", 0.55, 0.0),
+    CAMP_SITE("icon-campsite", "#734a08", "Camping", DEFAULT_BASE_SCORE, 0.0),
+    ATTRACTION("icon-star", "#ffb800", "Other", 0.55, 0.0),
+    ARTWORK("icon-artwork", "#ffb800", "Other", DEFAULT_BASE_SCORE, 0.0),
+    ALPINE_HUT("icon-alpinehut", "#734a08", "Camping", 0.55, 0.0),
+    HIGHWAY_CYCLEWAY("icon-bike", "black", "Bicycle", DEFAULT_BASE_SCORE, 0.0),
+    HIGHWAY_FOOT("icon-hike", "black", "Hiking", DEFAULT_BASE_SCORE, 0.0),
+    HIGHWAY_TRACK("icon-four-by-four", "black", "4x4", DEFAULT_BASE_SCORE, 0.0),
+    WORSHIP_JEWISH("icon-synagogue", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    WORSHIP_CHRISTIAN("icon-church", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    WORSHIP_MUSLIM("icon-mosque", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    WORSHIP_OTHER("icon-holy-place", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    INATURE("icon-inature", "#116C00", "iNature", DEFAULT_BASE_SCORE, 0.0),
+    FALLBACK_HISTORIC("icon-search", "black", "Other", 0.55, 0.0),
+    FALLBACK_WATER("icon-search", "black", "Other", 0.30, 0.0),
+    FALLBACK("icon-search", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+
+    NONICON_GENERIC("icon-search", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    NONICON_STATION("icon-bus-stop", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    NONICON_RIDGE("icon-peak", "black", "Other", DEFAULT_BASE_SCORE, 0.0),
+    NONICON_MTB("icon-bike", "green", "Bicycle", DEFAULT_BASE_SCORE, 0.0),
+    NONICON_FOREST("icon-tree", "#008000", "Other", DEFAULT_BASE_SCORE, 0.0),
+    NONICON_WIKIPEDIA("icon-wikipedia-w", "black", "Wikipedia", DEFAULT_BASE_SCORE, 0.0);
+
+    final String icon;
+    final String color;
+    final String poiCategory;
+    final double baseScore;
+    final double elevationWeight;
+
+    Category(String icon, String color, String poiCategory, double baseScore, double elevationWeight) {
+      this.icon = icon;
+      this.color = color;
+      this.poiCategory = poiCategory;
+      this.baseScore = baseScore;
+      this.elevationWeight = elevationWeight;
+    }
+  }
+
+  private OsmFeatureClassifier() {}
+
+  static Category classify(WithTags f) {
+    String boundary = f.getString("boundary");
+    if ("protected_area".equals(boundary) || "national_park".equals(boundary)) {
+      return Category.NATURE_RESERVE;
+    }
+    if ("nature_reserve".equals(f.getString("leisure"))) {
+      return Category.PROTECTED_NODE;
+    }
+
+    String route = f.getString("route");
+    if (route != null) {
+      switch (route) {
+        case "hiking":
+        case "foot":
+          return Category.ROUTE_HIKING;
+        case "bicycle":
+        case "mtb":
+          return Category.ROUTE_BICYCLE;
+        case "road":
+          if ("yes".equals(f.getString("scenic"))) {
+            return Category.ROUTE_4X4;
+          }
+      }
+    }
+
+    String historic = f.getString("historic");
+    if (historic != null) {
+      switch (historic) {
+        case "ruins":
+          return Category.HISTORIC_RUINS;
+        case "archaeological_site":
+          return Category.HISTORIC_ARCHAEOLOGICAL;
+        case "memorial":
+        case "monument":
+          return Category.HISTORIC_MEMORIAL;
+        case "tomb":
+          return Category.HISTORIC_TOMB;
+      }
+    }
+
+    if ("picnic_table".equals(f.getString("leisure"))
+        || "picnic_site".equals(f.getString("tourism"))
+        || "picnic".equals(f.getString("amenity"))) {
+      return Category.PICNIC;
+    }
+
+    String natural = f.getString("natural");
+    if (natural != null) {
+      switch (natural) {
+        case "cave_entrance":
+          return Category.NATURAL_CAVE;
+        case "spring":
+          return Category.NATURAL_SPRING;
+        case "tree":
+          return Category.NATURAL_TREE;
+        case "flowers":
+          return Category.NATURAL_FLOWERS;
+        case "waterhole":
+          return Category.NATURAL_WATERHOLE;
+        case "peak":
+        case "volcano":
+          return Category.PEAK;
+        case "ridge":
+          return Category.RIDGE;
+      }
+    }
+
+    String water = f.getString("water");
+    if ("reservoir".equals(water) || "pond".equals(water) || "lake".equals(water)
+        || "stream_pool".equals(water)) {
+      return Category.WATER_BODY;
+    }
+
+    String manMade = f.getString("man_made");
+    if (manMade != null) {
+      switch (manMade) {
+        case "water_well":
+          return Category.MAN_MADE_WATER_WELL;
+        case "cistern":
+          return Category.MAN_MADE_CISTERN;
+      }
+    }
+
+    String waterway = f.getString("waterway");
+    if ("waterfall".equals(waterway)) {
+      return Category.WATERFALL;
+    }
+
+    if ("waterway".equals(f.getString("type"))) {
+      return Category.WATERWAY_RELATION;
+    }
+
+    String place = f.getString("place");
+    if (place != null) {
+      if (place.isBlank()) {
+        return Category.PLACE_BLANK;
+      }
+      switch (place) {
+        case "city":
+          return Category.PLACE_CITY;
+        case "town":
+          return Category.PLACE_TOWN;
+        case "village":
+          return Category.PLACE_VILLAGE;
+        case "hamlet":
+          return Category.PLACE_HAMLET;
+        default:
+          return Category.PLACE_OTHER;
+      }
+    }
+
+    String tourism = f.getString("tourism");
+    if (tourism != null) {
+      switch (tourism) {
+        case "viewpoint":
+          return Category.VIEWPOINT;
+        case "camp_site":
+          return Category.CAMP_SITE;
+        case "attraction":
+          return Category.ATTRACTION;
+        case "artwork":
+          return Category.ARTWORK;
+        case "alpine_hut":
+          return Category.ALPINE_HUT;
+      }
+    }
+
+    String highway = f.getString("highway");
+    if (highway != null) {
+      switch (highway) {
+        case "cycleway":
+          return Category.HIGHWAY_CYCLEWAY;
+        case "footway":
+        case "path":
+          return Category.HIGHWAY_FOOT;
+        case "track":
+          return Category.HIGHWAY_TRACK;
+      }
+    }
+
+    String amenity = f.getString("amenity");
+    if ("place_of_worship".equals(amenity) || "monastery".equals(amenity)) {
+      String religion = f.getString("religion") != null ? f.getString("religion") : "";
+      switch (religion) {
+        case "jewish":
+          return Category.WORSHIP_JEWISH;
+        case "christian":
+          return Category.WORSHIP_CHRISTIAN;
+        case "muslim":
+          return Category.WORSHIP_MUSLIM;
+        default:
+          return Category.WORSHIP_OTHER;
+      }
+    }
+
+    if (f.getString("ref:IL:inature") != null) {
+      return Category.INATURE;
+    }
+
+    if (historic != null) {
+      return Category.FALLBACK_HISTORIC;
+    }
+    if ("hot_spring".equals(natural) || waterway != null) {
+      return Category.FALLBACK_WATER;
+    }
+    return Category.FALLBACK;
+  }
+
+  static Category classifyNonIcon(WithTags f) {
+    Category c = null;
+    if (f.hasTag("amenity", "place_of_worship") || f.hasTag("natural", "valley")) {
+      c = Category.NONICON_GENERIC;
+    }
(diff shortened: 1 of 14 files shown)


sergeytkachenko:
Follow-up from the code review: the order-dependent `ProminenceCalculator.baseScore()` (first-match precedence) is deferred and tracked separately in #83. Current behavior is kept in this PR.

sergeytkachenko:
@HarelM Please take a look

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -98,6 +100,67 @@ private void convertTagsToDocument(PointDocument pointDocument, WithTags feature
...
+        feature.getString("historic"),
+        feature.getString("waterway"),
+        ele, hasImage, hasWebsite, hasWikidata, qrankRaw);
+
+    pointDocument.poiProminence = score.prominence();
+  }
+
+  private static final Pattern FIRST_NUMBER_PATTERN = Pattern.compile("-?\\d[\\d ,']*(?:\\.\\d+)?");
+
+  static double parseFirstNumber(String raw) {
```
HarelM: I tend to think this is an overkill. Population should be an integer and if it has "," or "." simply remove them before parsing the string... Am I missing something? What other values can be in population?

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -571,7 +634,7 @@ private float normalizeArea(double areaM) {
     return (float) Math.max(0.0, Math.min(1.0, norm));
   }
 
-  private void addNonIconFeaturesToElasricseach(SourceFeature feature) throws GeometryException {
+  void addNonIconFeaturesToElasricseach(SourceFeature feature) throws GeometryException {
```
HarelM: Why was the private modifier removed? This should not be part of the public API of this class. If you think this should be in another class then move it there...

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -641,7 +705,12 @@ private void addNonIconFeaturesToElasricseach(SourceFeature feature) throws Geom
     insertPointToElasticsearch(pointDocument, docId);
   }
 
+  static float prominenceOrFloor(Float prominence) {
```
HarelM: This should be part of the calculator, not part of this class.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -641,7 +705,12 @@ private void addNonIconFeaturesToElasricseach(SourceFeature feature) throws Geom
...
+    return prominence == null ? (float) ProminenceCalculator.PROMINENCE_FLOOR : prominence;
+  }
+
   private void insertPointToElasticsearch(PointDocument pointDocument, String docId) {
+    pointDocument.poiProminence = prominenceOrFloor(pointDocument.poiProminence);
```
HarelM: There's a place where this is updated, let's try and keep it in one place, instead of adding it here as a safeguard.

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,71 @@
...
+
+    double meta = clamp01(0.40 * (hasImage ? 1 : 0) + 0.35 * (hasWebsite ? 1 : 0) + 0.25 * (hasWikidata ? 1 : 0));
+
+    double prominence = clamp01(PROMINENCE_FLOOR + WEIGHT_BASE * base + WEIGHT_QRANK * qrankNormalized + WEIGHT_META * meta);
+
+    return new ProminenceScore((float) prominence, (float) base, (float) qrankNormalized, (float) meta,
+        (float) eleNormalized, qrankRaw);
+  }
+
+  private static double baseScore(String natural, String place, String boundary, String tourism,
```
HarelM: As written in a previous PR review, this doesn't look like it captures all the stuff that the search is indexing in terms of OSM entities, so I think it would be better to try and extract the logic of the iconColorCategory method into a separate file and add for each icon-color-category its score, this way we can easily see that if we decide to support a new type of OSM tag we make sure we add the relevant scoring to it.

src/main/java/il/org/osm/israelhiking/QRankLookup.java
```diff
@@ -0,0 +1,105 @@
...
+  static QRankLookup empty() {
+    return new QRankLookup(new LongIntHashMap(0));
+  }
+
+  static QRankLookup load(Path csvGz) {
+    if (csvGz == null) {
+      LOGGER.info("QRankLookup: no QRank file configured — running with an empty lookup (lookups return 0)");
+      return empty();
+    }
+    if (!Files.isRegularFile(csvGz)) {
```
HarelM: Can we avoid this check? I don't see how we can cause this. The idea is simply to configure this correctly and the readme should indicate how to do that.

src/main/java/il/org/osm/israelhiking/QRankLookup.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+    long startMs = System.currentTimeMillis();
+    LongIntHashMap map = new LongIntHashMap(estimateRowCount(csvGz));
+    long rows = 0;
+    try (InputStream fis = Files.newInputStream(csvGz);
+        GZIPInputStream gz = new GZIPInputStream(fis, 1 << 16);
+        BufferedReader br = new BufferedReader(new InputStreamReader(gz, StandardCharsets.UTF_8), 1 << 20)) {
+      String line = br.readLine();
+      while ((line = br.readLine()) != null) {
+        int comma = line.indexOf(',');
```
HarelM: Can we use a CSV parser library instead of this code? I prefer to use something that's battle tested. I know how tempting it is to write a CSV parser...

src/test/java/il/org/osm/israelhiking/InsertProminenceFloorTest.java
HarelM, line 1: All the tests written here have no value, sorry, please remove them all. Maybe there's a single test that's meaningfull here, and in any case this should be part of the prominence calculator.

src/test/java/il/org/osm/israelhiking/NonIconFeatureProminenceTest.java
HarelM, line 1: I don't see value in these tests as well. Please remove them. Exporting some methods to a separate OSM tags helper class will allow better ways to test things in a meaningful way I believe.

src/test/java/il/org/osm/israelhiking/ParseFirstNumberTest.java
HarelM, line 1: I can now see which values parseFirstNumber should be handling, but I'm not sure both elevation and population should use the same parser.
if you want to compare elevation of different peaks, while some are in meters and some are in feet you should do a different parsing.
For population, you should parse to int instead of double.
So I would recommend extracting this method to a different class (like the OSM tags helper or an OSM number parser) and have two methods to handle this differently.
The tests should be kept as they show which values this parser supports.

src/test/java/il/org/osm/israelhiking/QRankLookupTest.java
```diff
@@ -0,0 +1,140 @@
...
+        var lookup = QRankLookup.load(dir.resolve("does-not-exist.csv.gz"));
+        assertEquals(0, lookup.size());
+    }
+
+    @Test
+    public void parsesGzippedCsvAndLooksUpByWikidata(@TempDir Path dir) throws IOException {
+        Path gz = dir.resolve("qrank.csv.gz");
+        try (OutputStream os = Files.newOutputStream(gz);
+                GZIPOutputStream gzos = new GZIPOutputStream(os)) {
+            gzos.write("Entity,QRank\nQ665321,359540\nQ121211166,34\nQ42,1000000\n"
```
HarelM: Are you use this is the format? In the following readme the there are 3 columns, not 2:
https://github.com/brawer/wikidata-qrank/blob/main/README.md

src/test/java/il/org/osm/israelhiking/QRankLookupTest.java
```diff
@@ -0,0 +1,140 @@
...
+        assertEquals(3, lookup.size());
+        assertEquals(359540, lookup.qrankFor("Q665321"));
+        assertEquals(34, lookup.qrankFor("Q121211166"));
+        assertEquals(1000000, lookup.qrankFor("Q42"));
+    }
+
+    @Test
+    public void multiValueWikidataUsesFirstQid(@TempDir Path dir) throws IOException {
+        Path gz = dir.resolve("qrank.csv.gz");
+        try (OutputStream os = Files.newOutputStream(gz);
```
HarelM: Consider extracting this repeated file write lines into a method and simply pass the relevant string to store in the file.

src/test/java/il/org/osm/israelhiking/QRankLookupTest.java
HarelM, line 1: I think some tests here can be "collapsed" - i.e. all the valid csv tests for different queries can be placed in the same test and they load the same "file".
Mostly nit-picking. The tests here are good in general.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Reverted to `private` — it should not be part of the class's public API. (The visibility was only widened for a test, which has since been removed.)

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Split this out into `OsmNumberParser.parsePopulation` (returns `int`), and it now strips `,`, `.`, `'` and grouping spaces before parsing, as you suggested (`1.234` → `1234`, `12'345` → `12345`). Values seen in OSM `population`: plain integers, thousands separators (`,`/`.`/`'`/space), and trailing noise like `5000 (2011)` — we take the first number run and strip separators. Elevation is parsed separately (it needs ft→m, which population doesn't).

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Removed `prominenceOrFloor` entirely. The floor now lives only in `ProminenceCalculator` (`PROMINENCE_FLOOR`, baked into `compute()`), so it is single-sourced.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Removed the insert-time net. One nuance worth flagging: I added a data-level default `PointDocument.poiProminence = PROMINENCE_FLOOR` (referencing the single constant, not a literal). Reason — with `@JsonInclude(NON_NULL)` a null prominence is omitted from the doc, and the query-side `field_value_factor missing:1.0` would then rank that feature at the **top**, which is worse than the floor. This keeps the floor value single-sourced while guaranteeing every document carries one, without an imperative safeguard at insert time. Happy to revisit if you'd prefer to handle it differently.

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
sergeytkachenko: Done — extracted `OsmFeatureClassifier`: a `Category` enum where each constant carries `icon`/`color`/`poiCategory` **and** a base score; `classify()`/`classifyNonIcon()` are the single source of truth, and `setIconColorCategory`/`addNonIcon` are now thin appliers. Adding a new OSM tag is one `Category` (with its score) in one place. Note: this extraction is behavior-preserving — every newly-distinguished category currently keeps the old `0.25` default; differentiated per-category scores change ranking, so I'd like to land them in a follow-up with an old-vs-new diff for your sign-off.

src/main/java/il/org/osm/israelhiking/QRankLookup.java
sergeytkachenko: Removed the explicit `isRegularFile` check. A missing/unreadable configured path still fails the build fast (the read attempt throws), so we don't silently floor every feature, but without the extra `stat()`. The README documents how to set `--qrank-path` (and that omitting it builds without QRank).

src/main/java/il/org/osm/israelhiking/QRankLookup.java
sergeytkachenko: Kept the hand-rolled parser, reasoning: the published file is a fixed, quote-free 2-column CSV (verified — see the QRankLookupTest thread) on a ~7M-row / ~363 MB hot path; the index-based parse (`indexOf(',')` + `Long.parseLong(line, 1, comma)`) avoids per-row allocation, and no CSV library is currently a dependency. I'd keep it for this trivial, stable format, but if you'd prefer something battle-tested I'll switch to `univocity-parsers` — your call.

src/test/java/il/org/osm/israelhiking/InsertProminenceFloorTest.java
sergeytkachenko: Removed this test file.

src/test/java/il/org/osm/israelhiking/NonIconFeatureProminenceTest.java
sergeytkachenko: Removed it; classification is now unit-tested directly via `OsmFeatureClassifierTest`. I did add one lean `PlanetSearchProfileIntegrationTest` that drives the **public** `processFeature` (no visibility relaxation) to guard that an indexed document actually carries a computed prominence + a fallback population — removing this test plus the insert-time net had otherwise left that end-to-end invariant untested. If you'd rather drop it too, I will.

src/test/java/il/org/osm/israelhiking/ParseFirstNumberTest.java
sergeytkachenko: Done — split into `OsmNumberParser` with `parsePopulation` (integer) and `parseElevation` (`double`, converts `ft`/`feet`/`'` → metres, which fixes feet-tagged peaks that were previously scored as metres). The documenting cases are kept, migrated into `OsmNumberParserTest` (split into population vs elevation, plus the new feet cases).

src/test/java/il/org/osm/israelhiking/QRankLookupTest.java
sergeytkachenko: Verified the published file is **2 columns**. I fetched the live `https://qrank.toolforge.org/download/qrank.csv.gz`: the header is `Entity,QRank` and rows look like `Q178995,559093819`. The `brawer/wikidata-qrank` README doesn't specify the download layout; toolforge serves the 2-column CSV, so the fixtures match the real file.

src/test/java/il/org/osm/israelhiking/QRankLookupTest.java
sergeytkachenko: Done — extracted a `writeGzip(dir, content)` helper used by all the gzip-producing tests.

src/test/java/il/org/osm/israelhiking/QRankLookupTest.java
sergeytkachenko: Done — collapsed the valid-lookup cases into a single `looksUpVariousWikidataFormsFromOneFile` that loads one fixture and asserts the different queries against it.

src/main/java/il/org/osm/israelhiking/OsmFeatureClassifier.java
```diff
@@ -0,0 +1,274 @@
+package il.org.osm.israelhiking;
+
+import com.onthegomap.planetiler.reader.WithTags;
+
+final class OsmFeatureClassifier {
+
+  enum Category {
+    NATURE_RESERVE("icon-leaf", "#008000", "Other", 0.80, 0.0),
+    PROTECTED_NODE("icon-leaf", "#008000", "Other", 0.25, 0.0),
```
HarelM: Is 0.25 a default low value? If so, please move it to a constant maybe?
I'm not sure, I'm still trying to figure out what happens if I need to change these values...

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -98,6 +98,50 @@ private void convertTagsToDocument(PointDocument pointDocument, WithTags feature
...
+
+  private void setProminence(PointDocument pointDocument, WithTags feature,
+      OsmFeatureClassifier.Category category) {
+    long qrankRaw = this.context.qrankLookup().qrankFor(pointDocument.wikidata);
+    double ele = OsmNumberParser.parseElevation(feature.getString("ele")).orElse(Double.NaN);
+    boolean hasImage = pointDocument.image != null || pointDocument.wikimedia_commons != null;
+    boolean hasWebsite = pointDocument.website != null;
+    boolean hasWikidata = pointDocument.wikidata != null;
+
+    ProminenceCalculator.ProminenceScore score = ProminenceCalculator.compute(
```
HarelM: Why returning a object and the calling it's prominence makes sense? Why not simply return the prominence number?

src/main/java/il/org/osm/israelhiking/QRankLookup.java
HarelM: Let's move to `univocity-parsers`, I've seen too many CSV issue in the past to know that I prefer a battle tested parser.

src/test/java/il/org/osm/israelhiking/PlanetSearchProfileIntegrationTest.java
HarelM, line 1: While I understand the appeal of these tests, I don't think they should exist this way.
If there's some logic that need to be extracted from the profile class in order to test it, I suggest adding the relevant file method to a different class.
Or completely remove these...

src/main/java/il/org/osm/israelhiking/QRankLookup.java
```diff
@@ -0,0 +1,99 @@
+package il.org.osm.israelhiking;
+
+import com.carrotsearch.hppc.LongIntHashMap;
```
HarelM, line 3: I saw that this was added in recent commit, doesn't planetiler have similar objects that can be used so there's no need to add a dependency just for this?

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
```diff
@@ -0,0 +1,40 @@
...
+    double eleNormalized = Double.isNaN(ele) ? 0.0 : clamp01(Math.log1p(Math.max(0, ele)) / Math.log1p(MAX_ELE_M));
+    double base = clamp01(category.baseScore + category.elevationWeight * eleNormalized);
+
+    double qrankNormalized = (!hasWikidata || qrankRaw <= 1) ? 0.0 : clamp01(Math.log(qrankRaw) / Math.log(QRANK_SCALE));
+
+    double meta = clamp01(0.40 * (hasImage ? 1 : 0) + 0.35 * (hasWebsite ? 1 : 0) + 0.25 * (hasWikidata ? 1 : 0));
+
+    double prominence = clamp01(PROMINENCE_FLOOR + WEIGHT_BASE * base + WEIGHT_QRANK * qrankNormalized + WEIGHT_META * meta);
+
+    return new ProminenceScore((float) prominence, (float) base, (float) qrankNormalized, (float) meta,
```
HarelM: Consider just returning the score and avoiding the need to define this object as it is not being used outside this method.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -98,6 +98,50 @@ private void convertTagsToDocument(PointDocument pointDocument, WithTags feature
     if (feature.hasTag("intermittent", "yes")) {
       pointDocument.intermittent = true;
     }
+    enrichRankingSignals(pointDocument, feature);
+  }
+
+  private void enrichRankingSignals(PointDocument pointDocument, WithTags feature) {
```
HarelM: I think this method can be removed as it is a one-liner...

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
```diff
@@ -632,6 +640,8 @@ private void addNonIconFeaturesToElasricseach(SourceFeature feature) throws Geom
     pointDocument.wikidata = feature.getString("wikidata");
     pointDocument.image = feature.getString("image");
     pointDocument.wikimedia_commons = feature.getString("wikimedia_commons");
+    pointDocument.website = feature.getString("website");
```
HarelM: I see that this looks like a duplication of convertTagsToDocument, can it be used here instead of repeating a substantial part of it here?
This is probably a mistake I made, but it's worth fixing it if possible now that we change this code...

HarelM:
I've added a few last comments, this should be ready to merge soon.

src/main/java/il/org/osm/israelhiking/OsmFeatureClassifier.java
sergeytkachenko: Yes — 0.25 is the default low base score. Extracted it to `DEFAULT_BASE_SCORE` in `OsmFeatureClassifier`; the special-cased categories keep their explicit values, so changing the default is now a one-liner.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Agreed — `ProminenceCalculator.compute` now returns the `float` prominence directly; the wrapper object is gone.

src/main/java/il/org/osm/israelhiking/QRankLookup.java
sergeytkachenko: Done — `QRankLookup.load` now parses the gzipped CSV with `univocity-parsers` (`CsvParser`) instead of the hand-rolled loop.

src/test/java/il/org/osm/israelhiking/PlanetSearchProfileIntegrationTest.java
sergeytkachenko: Removed. The profile is no longer instantiated in tests; classification, parsing and scoring are covered by isolated unit tests (`OsmFeatureClassifierTest`, `OsmNumberParserTest`, `ProminenceCalculatorTest`).

src/main/java/il/org/osm/israelhiking/QRankLookup.java
sergeytkachenko: Right — switched to planetiler's `Hppc.newLongIntHashMap()` and dropped the explicit `hppc` dependency (now only transitive via planetiler-core).

src/main/java/il/org/osm/israelhiking/ProminenceCalculator.java
sergeytkachenko: Done — `compute` returns the score `float` directly; the `ProminenceScore` record is removed.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Done — inlined the two calls into `convertTagsToDocument` and removed the method.

src/main/java/il/org/osm/israelhiking/PlanetSearchProfile.java
sergeytkachenko: Good catch — `addNonIconFeaturesToElasricseach` now calls `convertTagsToDocument` instead of duplicating the tag-copy / enrich block.

sergeytkachenko:
@HarelM thanks for the thorough review. Pushed a commit applying all of the latest round:

- `DEFAULT_BASE_SCORE` constant for the 0.25 default base score
- `ProminenceCalculator.compute` returns the prominence `float` directly (record removed)
- QRank CSV parsed with `univocity-parsers`
- planetiler's `Hppc.newLongIntHashMap()`; explicit `hppc` dependency dropped
- removed the profile-instantiating integration test
- inlined `enrichRankingSignals`
- `addNonIconFeaturesToElasricseach` reuses `convertTagsToDocument`

The PR is squashed to a single commit and marked ready for review. Tests are green (172).

sergeytkachenko:
@HarelM the latest review round is fully addressed (see the commit above) and CI is green (Java + Docker + codecov all passing). Should be ready to merge — let me know if anything else is needed.

HarelM (approved):
Great work! Thanks!

HarelM:
I'll add the qrank file to our production env.
Meanwhile I think we can move on to improve the search query in Site repo, right?
Or do you have something here that's more important?

sergeytkachenko:
> I'll add the qrank file to our production env. Meanwhile I think we can move on to improve the search query in Site repo, right? Or do you have something here that's more important?

Yes, we can switch to https://github.com/IsraelHikingMap/Site/pull/2543.
But I'll check if we haven't forgotten anything in this repository.

HarelM:
Sure, we can keep the index robustness fix to after the other PR is merged and in production as it's not gating to this feature.

PR #92, opened 2026-07-15
Gold set audit: strict entity-identity criteria, prefix flag wired, aspirational tier

sergeytkachenko:
Refs IsraelHikingMap/Site#2509, follows up on #90 (gold-set part of https://github.com/IsraelHikingMap/planet-search/issues/90#issuecomment-4973968304).

## What

Audit of every failing relevance gold case against a full-planet index (23.8M points), with each proposed entity id verified against the live OSM API. The failures split into real scorer bugs vs broken pins, and the harness gains the vocabulary to tell them apart:

- **`expectedIds`** — the case passes only when a top hit IS one of the intended OSM entities (normalized across the `OSM_` prefix difference between the ES ids and the Site API ids). Strict: a wrong entity inside the radius no longer passes. This fixed pins that were physically impossible — X011b had every same-named entity 12+ km out with a 10 km radius; D018's closest exact match sits at 2.6 km with a 2 km radius — and X004a where the lakes and the village are both legitimate answers.
- **`expectedExactName`** — for multi-segment linear features (T001 Pacific Crest Trail has 70 same-named segments; distance-to-pin is meaningless): a top hit must carry exactly that name within the radius.
- **`prefix: true`** is now actually sent by the relevance runner (`&prefix=true`). RX01 "grand can" and R011 "long pe" turn green with the intended target at rank 1 (4 m / 11 m) — they were only red because the harness never exercised the edge-ngram path.
- **`tier: aspirational`** (NLP-2, natural-language query — no NL parsing exists) is skipped consistently by both runners; the relevance runner still executes the search and reports the current outcome in the skip message, so it stays visible without gating.
- Load-time validation covers the new fields (id format, null elements, blank names) — malformed gold entries fail fast with `invalid search case: <id>` instead of NPEing mid-run.
- `SearchCasesTest` (new): strict-criteria unit coverage with synthetic cases + a smoke load of the real gold files.

## Result

301/314 → **307/314 + 1 skipped** on the live scorer, and every remaining failure is a real scorer gap with a clean, OSM-verified target:

| class | cases |
|---|---|
| far-famous (exact name of a famous feature loses to nearby partial matches) | FF01, FF03, FF08, FF10 (the Grand Canyon NP report) |
| prefix ranking | RX04 "mesa v" |
| town outranks the same-named lake on the map center | G0046, G0078 |

That table is the tuning backlog for the next scoring iteration ("exact match should travel wider" — measured in the issue thread).

Unit 193/193, E2E green on the israel extract, gold measured through the stored `points_search` template on a full-planet index.


src/test/java/il/org/osm/israelhiking/SearchCasesTest.java
HarelM, line 1: Is this a unit test for the end to end test helper?

HarelM:
@sergeytkachenko I've just merged an initial commit to fold the bbox into points index. Just FYI.

src/test/java/il/org/osm/israelhiking/SearchCasesTest.java
sergeytkachenko: Yes — SearchCases (Case, load/validate, failure) is the shared pass/fail helper both runners drive: E2ETest feeds it hits from the ES search template, SearchRelevanceTest feeds it hits from the live /api/search. Now that failure() carries real logic (strict expectedIds entity matching with OSM_ prefix normalization, exact-name-within-radius for multi-segment trails, load-time validation of the new fields), it gets direct unit coverage with synthetic cases — the two integration tests only exercise the happy paths of whatever the current index returns. The last test is a smoke-load of the two real gold files so a malformed entry fails the unit gate, not a live run.

Also rebased on top of the bbox-fold commits (#93/#94) — 193/193 unit tests green after the rebase.

HarelM: No need to test the tests...

sergeytkachenko: Fair enough — removed. The strict criteria stay in SearchCases itself and every E2E run exercises them.

HarelM:
This PR is still in draft, let me know if this is ready for review and merge.

src/test/java/il/org/osm/israelhiking/SearchCases.java
```diff
@@ -36,10 +38,14 @@ public boolean hasCenter() {
     public boolean isPrefix() {
       return Boolean.TRUE.equals(prefix);
     }
+
+    public boolean isAspirational() {
+      return "aspirational".equals(tier);
```
HarelM: I'm not sure I fully understand this tier part, even after looking at the search cases json object. Can this be changed to something more intuitive?

src/test/java/il/org/osm/israelhiking/SearchCases.java
```diff
@@ -59,40 +65,86 @@ public static List<Case> load(String resource) throws Exception {
   }
 
   private static void validate(Case searchCase) {
+    boolean validIds = searchCase.expectedIds() == null
```
HarelM, line 68: Would this make sense to use the document id (which is already a 1-1 OSM mapping) instead of adding this logic here?
I.e. write in the json OSM_node_xxx as the expected IDs?

HarelM:
This looks good overall, I would like to reduce the logic as much as possible in the e2e tests, but I understand if this is the minimal logic possible right now.
Let me know when this is no longer a draft and we can merge. Thanks!

src/test/java/il/org/osm/israelhiking/SearchCases.java
sergeytkachenko: Renamed to a plain boolean: `"allowFailure": true` (CI-style semantics — the case still runs and its current outcome shows up in the skip message, it just does not gate). `tier` is gone.

src/test/java/il/org/osm/israelhiking/SearchCases.java
sergeytkachenko: Done — the json now carries the literal document ids (`OSM_node_xxx`). The one-line normalization has to stay, though: the E2E path compares against ES `_id` (`OSM_node_xxx`) but SearchRelevanceTest compares against the Site API, which strips the source prefix and returns `node_xxx` — the strip bridges the two consumers of the same gold file.

HarelM (approved):

HarelM:
This looks good over all, I'm not sure what intent is used for, I would suggest to simply charge the ID of each case to be a test description which describes what this test is actually testing and remove intent completely.

sergeytkachenko:
Done: `intent` is gone, and every case id is now a test description — `find-eiffel-tower`, `prefix-grand-can`, `far-famous-denali`, `town-must-not-outrank-rice-lake-on-center`, `pacific-crest-trail-any-segment` — so a red case names what actually broke without opening the json. Unit and E2E green. Ready for merge from my side.

HarelM:
Thanks!

PR #93, opened 2026-07-15
Fold bbox infomation into the point index

HarelM:
This will allow removing the need for an extra query for container name, and in general will remove the need for the bbox index.
This is the first phase of only adding the relevant data and updating the query.
It works in a way that the first ever run will not have a places indexes etc, but will write the container index, later on the next run will pick it up.
If this design will be problematic I'll think of a way to improve it...
- Resolves #65


HarelM:
CC: @sergeytkachenko, FYI, I'll be pushing this forward in order to improve search performance, hopefully.

HarelM:
Ok, let's have it run in production and see how it goes.

PR #97, opened 2026-07-17
Fix relation management for places

HarelM:
In case a node and a relation exist for the same place, use the relation ID with the relevant merged tags from the node.
Also used this feature to move some code around, especially code related to place logic.


src/main/java/il/org/osm/israelhiking/PlaceIndex.java
```diff
@@ -0,0 +1,235 @@
...
+
+import com.onthegomap.planetiler.reader.SourceFeature;
+import com.onthegomap.planetiler.reader.WithTags;
+import com.onthegomap.planetiler.reader.osm.OsmElement;
+import com.onthegomap.planetiler.reader.osm.OsmSourceFeature;
+
+/**
+ * Tracks how each place is represented across the OSM node / way / relation
+ * forms seen in the first pass, so the second pass can index every place exactly
+ * once under the ranking <b>relation &gt; node &gt; way</b>:
```
HarelM:
```suggestion
 * once under the ranking relation then node then way:
```

